PR #8, opened 2026-02-05
docs: @vertz/core implementation plan

viniciusdacal:
## Summary

- Complete implementation plan for `@vertz/core` — the runtime package rewrite
- Replaces the legacy decorator-based, Fastify-dependent implementation with a functional, zero-dependency approach
- Custom HTTP server using Web Standard APIs (Request/Response) with adapters for Node.js, Bun, and Edge/Workers
- Compiler-driven DI — no reflect-metadata, no runtime resolution
- Middleware execution via return values (no `next()`), state accumulation, exception short-circuit
- Three-layer immutability: TypeScript DeepReadonly + Object.freeze (prod) + Proxy with helpful messages (dev)
- Service lifecycle hooks: `onInit(deps) → state → methods(deps, state) → onDestroy(deps, state)`
- 12 TDD implementation phases from foundation types through full integration

## Test plan

- [ ] Review plan covers all patterns from `vertz-core-api-design.md`
- [ ] Verify lifecycle hooks design addresses dbService use case
- [ ] Confirm package structure aligns with `__tests__/` convention from compiler plan
- [ ] Validate that open items are tracked

### DIFF
--- a/plans/vertz-core-implementation.md
+++ b/plans/vertz-core-implementation.md
@@ -72,10 +72,8 @@ packages/core/
 │   │
 │   ├── env/
 │   │   ├── index.ts
-│   │   ├── env-loader.ts                  # .env file parser (zero deps)
-│   │   ├── env-validator.ts               # Validates env against @vertz/schema
+│   │   ├── env-validator.ts               # Validates env against @vertz/schema, native .env loading
 │   │   └── __tests__/
-│   │       ├── env-loader.test.ts
 │   │       └── env-validator.test.ts
 │   │
 │   ├── server/
@@ -147,12 +145,7 @@ packages/core/
 │   │       ├── deps-builder.test.ts
 │   │       └── ctx-builder.test.ts
 │   │
-│   └── testing/
-│       ├── index.ts                       # vertz.testing namespace
-│       ├── test-app.ts                   # createApp() test builder
-│       ├── test-request.ts               # Simulated request execution
-│       └── __tests__/
-│           └── test-app.test.ts
+│   └── (no testing/ — lives in @vertz/testing package)
 ```
 
 ---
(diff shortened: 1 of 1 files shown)


plans/vertz-core-implementation.md
```diff
@@ -0,0 +1,696 @@
...
+```typescript
+// src/middleware/middleware-runner.ts
+export async function runMiddlewareChain(
+  middlewares: ResolvedMiddleware[],
+  initialCtx: Record<string, unknown>,
+): Promise<Record<string, unknown>> {
+  let state: Record<string, unknown> = {};
+
+  for (const mw of middlewares) {
+    const ctx = { ...initialCtx, state: { ...state } };
```
viniciusdacal: wondering if using  Object.assign(initialCtx, { state }) here and object.assign where we update the state, could impact positively on performance. Meaning, that would avoid recreating objects over and over, right? and as this is in our internal code and we have 100% control, it wouldn't cause side-effects, or better saying, we know exactly the scope of this side-effect

plans/vertz-core-implementation.md
```diff
@@ -0,0 +1,696 @@
...
+- `deepFreeze()`: nested objects, arrays, no-op on primitives
+- `createImmutableProxy()`: set throws with helpful message, delete throws, nested proxy
+- `makeImmutable()`: picks strategy by `NODE_ENV`
+
+### Phase 3: Environment
+
+**Files:** `env/`
+
+- `.env` file parser: key=value, quotes, comments, blank lines
+- `createEnv()`: schema validation, process.env merge, formatted error on failure
```
viniciusdacal: as I mentioned before... Can't we use the native approaches?

plans/vertz-core-implementation.md
```diff
@@ -0,0 +1,848 @@
...
+}
+```
+
+Handles OPTIONS preflight → 204. Adds headers to actual responses. Supports origin allowlisting, credentials, exposed headers, max-age.
+
+### 12. Testing Support
+
+Testing support lives entirely inside `@vertz/core` at `src/testing/`. Users import `vertz` from `@vertz/core` and access `vertz.testing.createApp()` — no separate package needed. The existing `packages/testing/` in the monorepo is superseded by this approach and will be removed.
+
+`vertz.testing.createApp()` returns a builder with the same patterns as the testing design plan:
```
viniciusdacal: I still think we should have a separate package for testing... What's the advantage of having it in the same package? We intend to release all packages with linked versions.

PR #89, opened 2026-02-08
docs: integration tests design plan

viniciusdacal:
## Summary

- Add design plan for `packages/integration-tests/` — a new workspace package that validates the framework works end-to-end
- Update README Current Status section to reflect all four packages are now stable

## Design plan highlights

A Users + Todos app exercising every framework primitive:

| Area | What's tested |
|------|--------------|
| CRUD lifecycle | GET/POST/PUT/DELETE with in-memory stores, state across requests |
| Schema validation | Body/params rejection for invalid input |
| Middleware | Auth middleware with real handler + mocking via `createTestApp()` |
| Error handling | 404, 405, custom exceptions, error body format |
| CORS | Preflight, response headers |
| Multi-module DI | Cross-module service injection (todos → users) |

Key design decisions documented: factory functions for isolation, `createApp()` vs `createTestApp()` trade-offs, router-level inject for cross-module DI, explicit body validation.

Plan: [plans/integration-tests.md](./plans/integration-tests.md)

## Test plan

This is a plan-only PR — no code changes beyond the plan document and README update.

### DIFF
--- a/plans/integration-tests.md
+++ b/plans/integration-tests.md
@@ -1,20 +1,38 @@
-# Integration Tests Package
+# End-to-End Integration Tests
 
 ## Context
 
-All four Vertz packages (`schema`, `core`, `testing`, `compiler`) are stable. We need a comprehensive end-to-end integration test that exercises all framework primitives together, simulating a real app. This validates the framework works as a whole and runs in CI.
+All four Vertz packages (`schema`, `core`, `testing`, `compiler`) are stable. We need a true end-to-end test that proves the framework works as a real HTTP server — spin up a server, make real `fetch()` requests, validate responses. No test utilities — this simulates the actual developer experience.
+
+## Prerequisite: Body/Query/Headers Schema Validation
+
+**Gap:** The framework currently only auto-validates `params` schemas in `app-runner.ts`. Body, query, and headers schemas are passed to route config but only used for TypeScript type inference — no runtime validation.
+
+**Fix (separate PR, before this plan):** Extend `RouteEntry` and `buildHandler` in `@vertz/core` to validate body, query, and headers schemas the same way params are validated — call `schema.parse()` before building ctx, throw `BadRequestException` on failure. The same fix must be applied to `@vertz/testing`'s `test-app.ts`.
+
+After this fix, route schemas become runtime contracts, not just type hints:
+```typescript
+router.post('/', {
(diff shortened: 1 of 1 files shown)


plans/integration-tests.md
```diff
@@ -0,0 +1,200 @@
...
+
+## App Design: Users + Todos
+
+### Key design decisions
+
+1. **Factory functions for modules** — each call returns fresh module + services + in-memory store. Ensures test isolation.
+2. **`createApp()` for stateful CRUD tests** — its handler caches services, so in-memory state persists across requests within a test. `createTestApp()` re-instantiates services per request, so it can't test create-then-read flows.
+3. **`createTestApp()` for single-request tests** — validation, error handling, mocking, response schema checks.
+4. **Router-level inject for cross-module DI** — the framework resolves service inject at router level, not service level. Todos router injects `userService` (exported from users module) so handler can access `ctx.userService`.
+5. **Handlers validate body explicitly** — the framework only auto-validates `params` schemas. Body/query schemas provide TypeScript types but not runtime validation. Handlers call `schema.parse(ctx.body)` and throw on failure.
```
viniciusdacal: Not sure if I got this one. I was expecting that we validate and parse the request body in the same place we validate request params, headers and query (before building ctx). Am I wrong?

plans/integration-tests.md
```diff
@@ -0,0 +1,200 @@
+# Integration Tests Package
```
viniciusdacal: I was expecting and e2e test here. We could setup a service with crud endpoints, run inside a docker, and trigger requests, via curl or via fetch and the framework should respond accordingly, should validate schemas, parser params, query, body, etc... 

We could try to simulate how it would work in a serverless environment at this point.

So, I don't expect to see vertz/testing here...

I mean, the purpose of this test is simulating that devs will be able to run this framework.

plans/integration-tests.md
viniciusdacal: You're right — body, query, and headers should be auto-validated the same way params are. Today only `params` schema gets `.parse()` called in `app-runner.ts` (line 144). The other schemas are just used for TypeScript type inference.

Added a **Prerequisite** section to the plan: we'll fix this in a separate PR before the E2E tests. After the fix, route schemas become runtime contracts — the framework validates before building ctx, throws BadRequestException on failure. No manual `schema.parse()` in handlers.

PR #99, opened 2026-02-08
docs: @vertz/ui design — compiler-driven UI library

viniciusdacal:
## Summary

- Design plan for `@vertz/ui` — a compiler-driven UI library that replaces React and integrates natively into the Vertz stack
- Competitive analysis of Svelte 5, SolidJS, Qwik, Marko, HTMX, and Million.js with concrete lessons applied
- Three parallel research agents produced: competitive analysis, compiler/reactivity architecture, and fullstack integration design

## Key Design Decisions

**Reactivity**: Plain `let` variables become reactive via compiler transformation. No `useState`, no `createSignal`, no `$state`. Just `let count = 0`. The compiler detects usage in JSX and generates signal-based subscriptions.

**Valid TypeScript**: All files are `.tsx` — not `.svelte`, not `.vue`. Code compiles with `tsc` (just not reactive without the Vertz compiler). LLMs and IDEs work perfectly.

**No useEffect**: DOM subscriptions are compiler-generated. For explicit side effects: `watch(() => dep, callback)` + `onCleanup()`. That's the entire API.

**Atomic hydration**: Per-component hydration with `<script type="application/json">` props. Only interactive components ship JS. Static HTML stays static.

**Streaming SSR**: Out-of-order streaming with `<Suspense>` boundaries. Slow data doesn't block fast content.

**Native forms**: `FormData` + `@vertz/schema` validation via `form()`. No `useState` per field. Progressive enhancement — works without JS.

**End-to-end types**: Compiler generates `route-types.ts` from backend IR. Types flow from schema → route → API client → component props.

**~4.5KB runtime**: Signals + DOM helpers + lifecycle. Smaller than Preact.

## Documents

- `plans/ui-design.md` — Unified design plan (18 sections)
- `plans/ui-competitive-analysis.md` — Deep analysis of 6 frameworks

## Test plan

- [ ] Review design plan for completeness and alignment with manifesto
- [ ] Validate compiler transformation examples (before/after)
- [ ] Check that "one way to do things" principle holds for all concerns
- [ ] Verify atomic hydration architecture handles edge cases

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,17 +1,12 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@3.1.2/schema.json",
-  "changelog": [
-    "@changesets/changelog-github",
-    { "repo": "vertz-dev/vertz" }
-  ],
-  "commit": false,
-  "fixed": [
-    ["@vertz/schema", "@vertz/core", "@vertz/compiler", "@vertz/testing"]
-  ],
-  "linked": [],
   "access": "public",
   "baseBranch": "main",
-  "updateInternalDependencies": "patch",
+  "bumpVersionsWithWorkspaceProtocolOnly": true,
+  "changelog": ["@changesets/changelog-github", { "repo": "vertz-dev/vertz" }],
+  "commit": false,
+  "fixed": [["@vertz/schema", "@vertz/core", "@vertz/compiler", "@vertz/testing"]],
   "ignore": [],
-  "bumpVersionsWithWorkspaceProtocolOnly": true
+  "linked": [],
+  "updateInternalDependencies": "patch"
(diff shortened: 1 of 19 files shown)


plans/ui-design.md
```diff
@@ -0,0 +1,789 @@
...
+```tsx
+import { ref } from "@vertz/ui";
+
+function Canvas() {
+  const canvasRef = ref<HTMLCanvasElement>();
+
+  watch(() => {
+    const ctx = canvasRef.current?.getContext("2d");
+    // imperative canvas drawing...
+  });
```
viniciusdacal: when watch would be executed here? only on mount on every time something changes?

PR #107, opened 2026-02-08
docs: @vertz/codegen design plan

viniciusdacal:
## Summary

- Comprehensive design plan for `@vertz/codegen` — a standalone package that generates typed SDK clients and CLI clients from the compiler's `AppIR`
- Covers architecture (IR adapter, generator plugin interface, template system), two built-in generators (TypeScript SDK + CLI manifest), and 8 implementation phases (~240 estimated tests)
- Key decisions: tagged template literals (no Handlebars), separate `CodegenIR` adapter layer, options-bag API shape, `createClient()` function pattern, Biome post-processing

## Key Design Decisions

- **Generators are pure functions** — receive IR, return `GeneratedFile[]`, no filesystem I/O
- **CodegenIR adapter** — thin projection of `AppIR` containing only the API surface (modules, operations, schemas)
- **Tagged template literals** — zero dependencies, full TypeScript power, LLM-friendly
- **Options-bag pattern** — `api.users.get({ params: { id: "abc" } })` over positional args
- **CLI as manifest** — generate data, not a full CLI; generic runtime consumes it
- **`{ content, imports }` tuple pattern** — each emit function returns content + needed imports, assembled at file level

## Test plan

- [ ] Review architecture diagram and CodegenIR types
- [ ] Review SDK API shape examples (Section 7)
- [ ] Review implementation phases and scope estimates (Section 15-16)
- [ ] Verify alignment with existing compiler patterns (`BaseGenerator`, `OpenAPIGenerator`)

### DIFF
--- a/plans/codegen-design.md
+++ b/plans/codegen-design.md
@@ -479,21 +479,30 @@ my-sdk/                           — publishable package root (user-owned)
   src/                — user's custom utilities (optional, entirely user-owned)
     helpers.ts
     middleware.ts
-  generated/          — codegen-owned, fully overwritten on every run
-    client.ts         — typed SDK client using @vertz/fetch
-    types.ts          — all operation input/output types
-    types/
+  generated/          — codegen-owned, overwritten on generate
+    client.ts         — createClient() that composes all modules
+    modules/          — one file per module (incremental-friendly)
+      users.ts        — users namespace methods
+      billing.ts      — billing namespace methods
+    types/            — one file per module
+      users.ts        — types for users operations
+      billing.ts      — types for billing operations
+      shared.ts       — named schemas used across modules
       augment.ts      — type augmentation entry point (declare module)
     schemas.ts        — re-exports of @vertz/schema objects
     index.ts          — barrel export of all generated code
 ```
 
+This per-module file layout enables **incremental regeneration** during `vertz dev` — changing a route only rewrites the affected module file and its types, not the entire SDK (see Section 17).
+
 **Rules**:
-1. `generated/` is **fully overwritten** on every `vertz generate` — users must never edit files in there
+1. `generated/` is codegen-owned — users must never edit files in there
 2. Root `index.ts` is **scaffolded once** — if it already exists, codegen does not touch it
 3. Root `package.json` is **scaffolded once** — same rule, never overwritten
 4. The scaffolded `index.ts` re-exports everything from `generated/` by default, but users can add their own exports
 5. Users add custom utilities anywhere outside `generated/` (e.g., `src/helpers.ts`) and re-export from root `index.ts`
+6. `vertz generate` (CLI/CI) does a **full regeneration** — overwrites all files in `generated/`
+7. `vertz dev` (watch mode) does **incremental regeneration** — only rewrites changed files (see Section 17)
 
 **Scaffolded `index.ts`** (generated only if file doesn't exist):
 
@@ -511,9 +520,15 @@ export * from './generated';
 
 ```
 .vertz/
-  generated/          — codegen-owned, gitignored, fully overwritten
+  generated/          — codegen-owned, gitignored
     client.ts
-    types.ts
+    modules/
+      users.ts
+      billing.ts
+    types/
+      users.ts
+      billing.ts
+      shared.ts
     schemas.ts
     index.ts
 ```
@@ -772,7 +787,7 @@ sdk/
 
 ---
 
-## 8. CLI Generator
+## 9. CLI Generator
 
 The CLI generator produces a **command manifest** — a data structure describing all available commands, their arguments, and types. A `@vertz/cli-runtime` package consumes this manifest at runtime to produce a complete CLI.
 
@@ -884,7 +899,7 @@ cli.run(process.argv.slice(2));
 
 ---
 
-## 9. `@vertz/cli-runtime` — CLI Runtime Package
+## 10. `@vertz/cli-runtime` — CLI Runtime Package
 
 A separate, open-source package that combines a command manifest + common interactive flows to produce a complete CLI tool. This is a runtime dependency of generated CLIs, not a codegen dependency.
 
@@ -969,7 +984,7 @@ Auth configuration is derived from the API's OpenAPI security schemes (via `Code
 
 ---
 
-## 10. Type Augmentation for Customer-Specific Types
+## 11. Type Augmentation for Customer-Specific Types
 
 For SaaS platforms where customers define custom resource types, the generated SDK supports type augmentation via TypeScript's `declare module` pattern.
 
@@ -1043,7 +1058,7 @@ For most APIs, this is not needed. It's an advanced feature for SaaS platforms t
 
 ---
 
-## 11. Template System: Tagged Template Functions
+## 12. Template System: Tagged Template Functions
 
 ### Why Not Handlebars/EJS/Eta?
 
@@ -1135,7 +1150,7 @@ function renderImports(imports: Import[]): string {
 
 ---
 
-## 12. Shared Utilities vs. Per-Generator Code
+## 13. Shared Utilities vs. Per-Generator Code
 
 ### Shared (in `utils/`)
 
@@ -1162,7 +1177,7 @@ function streamingOperations(ir: CodegenIR): CodegenOperation[]
 
 ---
 
-## 13. Configuration
+## 14. Configuration
 
 ### Codegen Config (in `vertz.config.ts`)
 
@@ -1221,7 +1236,7 @@ type GeneratorName = 'typescript' | 'cli';
 
 ---
 
-## 14. Output Location
+## 15. Output Location
 
 Two modes based on configuration:
 
@@ -1266,7 +1281,7 @@ The `generated/` subfolder should have its own `.gitignore` with just a comment 
 
 ---
 
-## 15. Formatting Pipeline
+## 16. Formatting Pipeline
 
 All generated TypeScript files are post-processed with Biome (already in the project):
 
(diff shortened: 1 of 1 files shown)


viniciusdacal:
Yes, we should consider that the SDKs we are generating might be published to npm as public or private packages. Also, we should be able to build and generate them in CI, so if people don't want to publish them, they don't need to. For the CLI, it's okay to keep it like that, just generating a schema that could be consumed by something else.

I'm not concerned about consuming or generating a CLI right now, so we could consume it for the frontend or so, but I'm more concerned about being able to generate a CLI that somehow someone could build the CLI into an npm package and publish so other people could use. 


Imagine the use case: I'm building a SaaS. I'm defining the API. I'm making it with the open API spec so everybody could consume. I'm making an SDK available, but at the same time I'm also making available a CLI. If someone wanted to consume from the terminal using Claude Code or some other LLM that runs inside a terminal, they could just use the CLI, right?

That's the idea: we want to make this easy for everyone that's building the APIs with words to generate those CLIs, build them, and publish to npm. 


However, I completely agree that we could simply generate that metadata. At the same time, we could have a package or so. We suggested that we have a package for Fetch; we could have an extra package that we could open source that you can combine with the metadata, and then you output a complete CLI tool.
---

We also have to consider that we might want to add some customization to the CLI, and perhaps we would. I'm just thinking we could add some metadata to the endpoints when we want. For example, I'm going to tell our use case.

Let's say a user wants to make a call to an endpoint using the CLI, and then the CLI is expecting a workspace ID. The user didn't provide that workspace ID, and then we need to list the available workspaces for that user. For that, we are going to need to fetch the API tool to get the list of workspaces and then list them to the user. The user selects the workspace, and then we populate that workspace ID in the CLI.

The CLI should be able to handle this kind of flow, and this is a common flow that could happen. The user should be able to have a way to configure this when defining the API. When they define the API somehow, like a route, they could define this kind of behavior. It could be like a second file that we could put close to the root, but the user definitely should be able to configure this kind of behavior. 

In this open source package that builds the CLI, we could also put some common flows for authentication, for example, for the OAuth authentication, and for other common stuff that we usually need when building the CLI. 

---

An extra use case: this is less common, but it helps a lot when we are building very type-safe SDKs for each customer. Let's say I'm building a Sass, and inside this Sass we have something that we call Resource Type, and the user is able to define Resource Types for his account. We might have a collection of Resource Types per customer, right? We want to make it possible that when the user imports our SDK, somehow we could have a CLI or so that would generate type augmentation, or the user could just combine some config file that he defined locally with the client, the SDK that we generated. Then he would get those custom types, right? Like he defined a collection of resource types and, let's say, three resource types, and then when he's trying to use our SDK, we are inferring those resource types.

One of the possibilities of doing this is using our schema package. When we allow the user to define a configuration, under the hood we could be using the schema to infer the types, and then we can let the user combine that type when creating the SDK. Another way would be type augmentation, kind of the same way the user would define the configuration in a config file, like a TypeScript file. We would use our schema to infer the types and then generate a type augmentation file only for those specific types.

The only problem is that TypeScript doesn't seem to like type augmentation for simple types, like for enums, for example. For primitives, it doesn't seem that we are able to simply augment the types without overriding a full package, right? What I was doing in another project was generating a separate namespace, like a separate npm package actually. Then we were referring to that npm package inside our schema that we auto-generate with the SDK, and then, when we use type augmentation, we were overriding only that specific package. I'll try to share an example here. 


https://github.com/blimu-dev/blimu-ts/blob/main/packages/backend/src/schema.ts

see: `import type { ResourceType, EntitlementType, PlanType, UsageLimitType } from '@blimu/types';`

This is only because we later can do this and typescript works:

```ts
/**
 * Type Augmentation for @blimu/types and @blimu/backend
 *
 * This file augments the @blimu/types and @blimu/backend packages with union types
 * specific to your environment configuration.
 *
 * Types are automatically inferred from your blimu.config.ts file.
 * No regeneration needed when you update your config!
 *
 * Make sure to include this file in your tsconfig.json:
 * {
 *   "include": ["blimu-types.d.ts"]
 * }
 */

import type config from './blimu.config';
import type {
  InferResourceTypes,
  InferEntitlementTypes,
  InferPlanTypes,
  InferLimitTypes,
  InferUsageLimitTypes,
} from 'blimu';

declare module '@blimu/types' {
  /**
   * Resource types inferred from your Blimu configuration.
   */
  type ResourceType = InferResourceTypes<typeof config>;

  /**
   * Entitlement types inferred from your Blimu configuration.
   */
  type EntitlementType = InferEntitlementTypes<typeof config>;

  /**
   * Plan types inferred from your Blimu configuration.
   */
  type PlanType = InferPlanTypes<typeof config>;

  /**
   * Limit types inferred from your Blimu configuration.
   */
  type LimitType = InferLimitTypes<typeof config>;

  /**
   * Usage limit types inferred from your Blimu configuration.
   */
  type UsageLimitType = InferUsageLimitTypes<typeof config>;
}

declare module '@blimu/backend' {
  /**
   * Resource types inferred from your Blimu configuration.
   */
  type ResourceType = InferResourceTypes<typeof config>;

  /**
   * Entitlement types inferred from your Blimu configuration.
   */
  type EntitlementType = InferEntitlementTypes<typeof config>;

  /**
   * Plan types inferred from your Blimu configuration.
   */
  type PlanType = InferPlanTypes<typeof config>;

  /**
   * Limit types inferred from your Blimu configuration.
   */
  type LimitType = InferLimitTypes<typeof config>;

  /**
   * Usage limit types inferred from your Blimu configuration.
   */
  type UsageLimitType = InferUsageLimitTypes<typeof config>;
}
```

So my suggestion would be that, when we have components, because we are able to define the schema using our schema package, we are able to define an ID for the model. Even if it's a string or some primitive, we are able to define that ID. I think that, when we have those primitive components, we could perhaps put them in a separate package that is a soup package from our key package. I would like to show an example of what, in my head, would be ideal. 


Package is available as '@blimu/backend'

and those primitive types are referenced as '@blimu/backend/types'

Then, type augmentation could override with `declare module '@blimu/backend/types' {`

viniciusdacal:
Yes, we should consider that the SDKs we are generating might be published to npm as public or private packages. Also, we should be able to build and generate them in CI, so if people don't want to publish them, they don't need to. For the CLI, it's okay to keep it like that, just generating a schema that could be consumed by something else.

I'm not concerned about consuming or generating a CLI right now, so we could consume it for the frontend or so, but I'm more concerned about being able to generate a CLI that somehow someone could build the CLI into an npm package and publish so other people could use. 


Imagine the use case: I'm building a SaaS. I'm defining the API. I'm making it with the open API spec so everybody could consume. I'm making an SDK available, but at the same time I'm also making available a CLI. If someone wanted to consume from the terminal using Claude Code or some other LLM that runs inside a terminal, they could just use the CLI, right?

That's the idea: we want to make this easy for everyone that's building the APIs with words to generate those CLIs, build them, and publish to npm. 


However, I completely agree that we could simply generate that metadata. At the same time, we could have a package or so. We suggested that we have a package for Fetch; we could have an extra package that we could open source that you can combine with the metadata, and then you output a complete CLI tool.
---

We also have to consider that we might want to add some customization to the CLI, and perhaps we would. I'm just thinking we could add some metadata to the endpoints when we want. For example, I'm going to tell our use case.

Let's say a user wants to make a call to an endpoint using the CLI, and then the CLI is expecting a workspace ID. The user didn't provide that workspace ID, and then we need to list the available workspaces for that user. For that, we are going to need to fetch the API tool to get the list of workspaces and then list them to the user. The user selects the workspace, and then we populate that workspace ID in the CLI.

The CLI should be able to handle this kind of flow, and this is a common flow that could happen. The user should be able to have a way to configure this when defining the API. When they define the API somehow, like a route, they could define this kind of behavior. It could be like a second file that we could put close to the root, but the user definitely should be able to configure this kind of behavior. 

In this open source package that builds the CLI, we could also put some common flows for authentication, for example, for the OAuth authentication, and for other common stuff that we usually need when building the CLI. 

---

An extra use case: this is less common, but it helps a lot when we are building very type-safe SDKs for each customer. Let's say I'm building a Sass, and inside this Sass we have something that we call Resource Type, and the user is able to define Resource Types for his account. We might have a collection of Resource Types per customer, right? We want to make it possible that when the user imports our SDK, somehow we could have a CLI or so that would generate type augmentation, or the user could just combine some config file that he defined locally with the client, the SDK that we generated. Then he would get those custom types, right? Like he defined a collection of resource types and, let's say, three resource types, and then when he's trying to use our SDK, we are inferring those resource types.

One of the possibilities of doing this is using our schema package. When we allow the user to define a configuration, under the hood we could be using the schema to infer the types, and then we can let the user combine that type when creating the SDK. Another way would be type augmentation, kind of the same way the user would define the configuration in a config file, like a TypeScript file. We would use our schema to infer the types and then generate a type augmentation file only for those specific types.

The only problem is that TypeScript doesn't seem to like type augmentation for simple types, like for enums, for example. For primitives, it doesn't seem that we are able to simply augment the types without overriding a full package, right? What I was doing in another project was generating a separate namespace, like a separate npm package actually. Then we were referring to that npm package inside our schema that we auto-generate with the SDK, and then, when we use type augmentation, we were overriding only that specific package. I'll try to share an example here. 


https://github.com/blimu-dev/blimu-ts/blob/main/packages/backend/src/schema.ts

see: `import type { ResourceType, EntitlementType, PlanType, UsageLimitType } from '@blimu/types';`

This is only because we later can do this and typescript works:

```ts
/**
 * Type Augmentation for @blimu/types and @blimu/backend
 *
 * This file augments the @blimu/types and @blimu/backend packages with union types
 * specific to your environment configuration.
 *
 * Types are automatically inferred from your blimu.config.ts file.
 * No regeneration needed when you update your config!
 *
 * Make sure to include this file in your tsconfig.json:
 * {
 *   "include": ["blimu-types.d.ts"]
 * }
 */

import type config from './blimu.config';
import type {
  InferResourceTypes,
  InferEntitlementTypes,
  InferPlanTypes,
  InferLimitTypes,
  InferUsageLimitTypes,
} from 'blimu';

declare module '@blimu/types' {
  /**
   * Resource types inferred from your Blimu configuration.
   */
  type ResourceType = InferResourceTypes<typeof config>;

  /**
   * Entitlement types inferred from your Blimu configuration.
   */
  type EntitlementType = InferEntitlementTypes<typeof config>;

  /**
   * Plan types inferred from your Blimu configuration.
   */
  type PlanType = InferPlanTypes<typeof config>;

  /**
   * Limit types inferred from your Blimu configuration.
   */
  type LimitType = InferLimitTypes<typeof config>;

  /**
   * Usage limit types inferred from your Blimu configuration.
   */
  type UsageLimitType = InferUsageLimitTypes<typeof config>;
}

declare module '@blimu/backend' {
  /**
   * Resource types inferred from your Blimu configuration.
   */
  type ResourceType = InferResourceTypes<typeof config>;

  /**
   * Entitlement types inferred from your Blimu configuration.
   */
  type EntitlementType = InferEntitlementTypes<typeof config>;

  /**
   * Plan types inferred from your Blimu configuration.
   */
  type PlanType = InferPlanTypes<typeof config>;

  /**
   * Limit types inferred from your Blimu configuration.
   */
  type LimitType = InferLimitTypes<typeof config>;

  /**
   * Usage limit types inferred from your Blimu configuration.
   */
  type UsageLimitType = InferUsageLimitTypes<typeof config>;
}
```

So my suggestion would be that, when we have components, because we are able to define the schema using our schema package, we are able to define an ID for the model. Even if it's a string or some primitive, we are able to define that ID. I think that, when we have those primitive components, we could perhaps put them in a separate package that is a soup package from our key package. I would like to show an example of what, in my head, would be ideal. 


Package is available as '@blimu/backend'

and those primitive types are referenced as '@blimu/backend/types'

Then, type augmentation could override with `declare module '@blimu/backend/types' {`

Issue #446, opened 2026-02-18
design: Canvas Phase 2 — JSX-based canvas rendering


viniciusdacal:
## Description

Phase 2 of the canvas renderer introduces JSX-based declarative canvas rendering. This requires a design doc before implementation begins.

The spike recommended a **hybrid approach**: DOM for text/forms/accessibility, Canvas for performance-critical graphics zones. Phase 2 should define:

1. **JSX runtime** — How do we compile JSX to PixiJS display objects instead of DOM elements?
2. **Component model** — What does a canvas component look like? How does it compose?
3. **Hybrid boundary** — How does a developer embed a `<CanvasLayer>` inside DOM and vice versa?
4. **Event system** — How do pointer/click events work on canvas elements?
5. **Lifecycle** — Mount, update, unmount for canvas components
6. **Reconciliation** — Do we need a virtual scene graph diffing algorithm?

## Acceptance Criteria

- [ ] Design doc written at `plans/canvas-phase-2.md`
- [ ] API surface defined with examples
- [ ] josh reviews DX
- [ ] mike reviews architecture
- [ ] Hard problems from spike addressed (text rendering, accessibility, layout)
- [ ] Decision: custom JSX factory vs compiler plugin vs runtime adapter

## References

- Phase 1 design: `plans/canvas-phase-1.md`
- Spike findings: `spike/canvas-rendering-poc` branch
- Spike summary: kai's workspace `spike-canvas-summary.md`
- Hard problems: text rendering (red), accessibility (red), layout engine (yellow), DevTools (yellow)

viniciusdacal:
Closing — Canvas Phase 2 (JSX-based canvas rendering) is fully implemented.

Relevant commits:
- `172ae3b` plan: Canvas Phase 2 design doc — JSX-based canvas rendering (#446)
- `1c810a5` feat(ui-canvas): Phase 2 — JSX-based canvas rendering (#454)
- `a7a451f` fix(ui-canvas): address review findings — disposal, children, types, CanvasLayer

Full `@vertz/ui-canvas` package with JSX-to-PixiJS rendering (`jsxCanvas()`), canvas layers, conditional/list support, and comprehensive test suite.

PR #448, opened 2026-02-18
plan: Canvas Phase 2 design doc — JSX-based canvas rendering

viniciusdacal:
## Summary

Design doc for `@vertz/ui-canvas` Phase 2: declarative JSX-based canvas rendering with hybrid DOM/Canvas architecture.

- Defines JSX runtime (`jsxCanvas` factory), component model, and intrinsic canvas elements (`<Graphics>`, `<Sprite>`, `<Container>`, `<Text>`)
- Designs `<CanvasLayer>` bridge component for hybrid DOM/Canvas boundary
- Addresses hard problems: text rendering, accessibility, layout engine, DevTools
- Includes API surface, type flow map, E2E acceptance test, manifesto alignment

## Key Design Decisions

| Decision | Choice | Rationale |
|----------|--------|-----------|
| Reconciliation | **None** — direct signal bindings | O(1) updates per signal change, aligns with Vertz's fine-grained reactivity |
| JSX Runtime | **Custom `jsxCanvas()` factory** | Runtime approach, no compiler coupling, preserves future optimization path |
| DOM/Canvas Boundary | **`<CanvasLayer>` bridge component** | DOM component hosts PixiJS app, children rendered into `app.stage` |
| Text Rendering | **PixiJS Canvas2D backend** | Built-in, good enough for Phase 2. MSDF deferred. |
| Accessibility | **Hidden DOM mirror (opt-in)** | Industry standard (Figma, Google Docs). `accessible={true}` prop. |
| Layout Engine | **Deferred to Phase 3** | Absolute positioning only. Yoga/Taffy evaluation later. |

## Unknowns

1. Canvas JSX runtime activation mechanism — discussion-resolvable
2. Reactive list rendering in canvas — discussion-resolvable
3. Graphics redraw performance with `clear() + redraw` — **needs POC**

## Review Requested

- **josh** — DX review (component API, developer ergonomics)
- **mike** — Architecture review (JSX runtime, signal integration, reconciliation decision)
- **pm** — Scope review (Phase 2 boundaries, non-goals)

Addresses #446

### DIFF
--- a/plans/canvas-phase-2.md
+++ b/plans/canvas-phase-2.md
@@ -108,6 +108,22 @@ function Circle(props: CircleProps) {
 
 The `unwrap` utility resolves `T | (() => T)` to `T`. Inside a `draw` callback that runs within an `effect`, accessor calls are tracked automatically.
 
+`unwrap` is exported from `@vertz/ui-canvas` and is essential for custom canvas components:
+
+```typescript
+import { unwrap } from '@vertz/ui-canvas';
+
+// Type signature:
+function unwrap<T>(value: T | (() => T)): T;
+
+// Implementation:
+function unwrap<T>(value: T | (() => T)): T {
+  return typeof value === 'function' ? (value as () => T)() : value;
+}
+```
+
+In dev mode, `unwrap` logs a warning if called outside an `effect` context with an accessor argument — since the signal read won't be tracked, the value won't update reactively.
+
 ### Hybrid DOM/Canvas Boundary — `<CanvasLayer>`
 
 `<CanvasLayer>` is the bridge between DOM and Canvas worlds. It is a DOM component that mounts a PixiJS `Application` into the DOM and renders its children as PixiJS display objects.
@@ -168,7 +184,6 @@ Canvas elements support PixiJS pointer events via the same `on*` convention as D
 <Graphics
   x={100} y={100}
   draw={(g) => { g.rect(0, 0, 100, 50); g.fill(0x4ecdc4); }}
-  interactive={true}
   onPointerDown={(e: FederatedPointerEvent) => {
     console.log('clicked at', e.global.x, e.global.y);
   }}
@@ -176,6 +191,12 @@ Canvas elements support PixiJS pointer events via the same `on*` convention as D
   onPointerOut={() => { hovered.value = false; }}
 />
 ```
+
+**Auto-interactive:** When any `on*` event prop is present, the runtime automatically sets `displayObject.eventMode = 'static'` (equivalent to PixiJS `interactive = true`). Developers do NOT need to manually set `interactive={true}`. This matches DOM behavior where elements are always interactive — no opt-in needed.
+
+> **Resolved from review (josh):** The original design required explicit `interactive={true}` for events to work. Every DOM developer would forget this. Auto-setting interactive when `on*` props are present eliminates the footgun.
+
+If a developer explicitly sets `interactive={false}`, that takes precedence (opt-out). The `eventMode` prop is also exposed for advanced control (`'static'`, `'passive'`, `'dynamic'`, `'auto'`).
 
 Events flow through PixiJS's built-in `FederatedEventSystem`. The canvas JSX runtime calls `displayObject.on('pointerdown', handler)` — no custom event system needed. PixiJS v8's event system already provides bubbling, hit-testing, and propagation for display objects.
 
@@ -221,6 +242,73 @@ function AnimatedCircle(props: { x: () => number; y: () => number }) {
 
 **Unmount:** When a parent removes a child (e.g., conditional rendering or list change), the disposal scope runs all registered cleanups, the display object is removed from its parent container, and all bound effects are disposed.
 
+### `ref` Escape Hatch
+
+For cases where developers need direct access to the underlying PixiJS display object (custom PixiJS APIs, advanced hit areas, manual transformations), canvas elements accept a `ref` prop:
+
+```typescript
+import { Graphics } from '@vertz/ui-canvas';
+
+function CustomShape() {
+  let graphicsRef: PIXI.Graphics;
+
+  return (
+    <Graphics
+      ref={(g) => { graphicsRef = g; }}
+      x={0} y={0}
+      draw={(g) => { g.rect(0, 0, 100, 50); g.fill(0xff0000); }}
+    />
+  );
+}
+```
+
+The `ref` callback receives the raw PixiJS display object after creation but before children are processed. This is the same pattern as DOM `ref` in Vertz.
+
+### Built-in Convenience Components
+
+In addition to the core intrinsic elements, `@vertz/ui-canvas` exports convenience components for common shapes:
+
+```typescript
+import { Circle, Rect, Line, Ellipse } from '@vertz/ui-canvas';
+
+// These are thin wrappers around <Graphics> with typed props
+<Circle x={100} y={100} radius={50} fill={0xff6347} />
+<Rect x={200} y={150} width={120} height={80} fill={0x4ecdc4} stroke={0xffffff} strokeWidth={2} />
+<Line from={{ x: 0, y: 0 }} to={{ x: 200, y: 100 }} stroke={0xffffff} strokeWidth={1} />
+<Ellipse x={300} y={200} halfWidth={60} halfHeight={40} fill={0x9b59b6} />
+```
+
+These are implemented as canvas components (not new intrinsic types) — they compose `<Graphics>` internally with pre-built `draw` callbacks. This keeps the intrinsic element set small while providing the 80% use case out of the box.
+
+### Sprite Loading and Error Handling
+
+`<Sprite>` loads textures asynchronously via `PIXI.Assets.load()`:
+
+```typescript
+<Sprite x={100} y={100} texture="assets/player.png" />
+```
+
+**Loading behavior:**
+- The sprite is created immediately (added to scene graph) but is invisible until the texture loads.
+- When `PIXI.Assets.load(path)` resolves, `sprite.texture` is set and the sprite becomes visible.
+- If the same path is requested multiple times, PixiJS's built-in cache returns the cached texture.
+
+**Error behavior:**
+- If the texture fails to load, a warning is logged in dev mode.
+- The sprite remains in the scene graph with no texture (invisible). This prevents the scene from breaking due to a missing asset.
+- A future `onError` prop can be added if explicit error handling is needed — not in Phase 2 scope.
+
+### Nested `<CanvasLayer>` — Forbidden
+
+Nesting `<CanvasLayer>` inside another `<CanvasLayer>` is explicitly forbidden. The runtime throws an error:
+
+```
+Error: <CanvasLayer> cannot be nested inside another <CanvasLayer>.
+Use <Container> to group canvas elements.
+```
+
+Each `<CanvasLayer>` manages its own PixiJS `Application`. Nesting would create multiple applications with undefined parent-child relationships. Use `<Container>` for grouping within a single canvas.
+
 ---
 
 ## Architecture Decision: No Virtual Scene Graph Reconciliation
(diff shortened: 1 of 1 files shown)


viniciusdacal:
## Architecture Review — Canvas Phase 2 Design Doc

**Reviewer:** mike (vertz-tech-lead)
**Date:** 2026-02-18

### Overall Assessment: Approve with Comments

The design is architecturally sound and aligns well with Vertz's fine-grained reactivity model. The no-reconciliation decision is correct and the signal integration approach is proven by Phase 1. There are, however, several areas that need clarification or correction before implementation begins. I am flagging two issues that could cause real bugs at runtime if not addressed in the design.

---

### Architecture Soundness

The overall architecture is solid. The key decisions are correct:

1. **Direct signal binding over virtual scene graph reconciliation** — This is the right call. Vertz's `effect()` system gives us O(1) updates per signal change. A reconciliation layer would add complexity and overhead with zero benefit given our reactive model. The design correctly identifies that this is consistent with how DOM rendering works in Vertz.

2. **Runtime JSX factory over compiler plugin** — Correct for Phase 2. The spike proved 1000+ interactive elements at 60fps, and the runtime factory preserves the option for compiler optimization later without locking us into premature complexity.

3. **`<CanvasLayer>` as the DOM-to-Canvas bridge** — Clean boundary. The pattern of a DOM component that mounts a PixiJS Application and renders children into the canvas scene graph is well-established (React Three Fiber's `<Canvas>` uses the same pattern).

4. **Sharing the reactive graph across the DOM/Canvas boundary** — This is a major strength. Signals created in DOM scope are readable in canvas components because the tracking system is global (there is one `currentSubscriber` in `tracking.ts`). This is not a feature that needs to be built — it falls out naturally from the existing architecture.

---

### Signal Integration Analysis

I have read the signal implementation in `packages/ui/src/runtime/signal.ts` and the disposal system in `disposal.ts` carefully. The design's approach is compatible, but there are nuances that need attention.

#### Auto-tracking works correctly for reactive props

The design proposes:
```typescript
effect(() => {
  displayObject.x = (value as () => unknown)();
})
```

This is correct. When the factory encounters a function-typed prop, it creates an `effect()` that calls the accessor. Inside the effect, `getSubscriber()` is set to the effect's subscriber, so any signal read via `.value` is tracked. When the signal changes, the effect re-runs and updates the display object property. This is identical to how `__attr()` works for DOM attributes.

#### The `draw` callback tracking is correct but has a subtle implication

The design proposes:
```typescript
effect(() => {
  displayObject.clear();
  (value as DrawFn)(displayObject);
})
```

This correctly tracks any signal reads inside the draw callback. However, the design should explicitly document that **all signal reads inside `draw` become dependencies of a single effect**. If the draw callback reads 5 different signals, changing any one of them triggers a full `clear() + redraw`. There is no way to partially redraw — the entire Graphics object is redrawn.

This is acceptable for Phase 2, but the documentation should be explicit so developers know that complex `draw` callbacks with many signal dependencies will redraw fully on any dependency change. The mitigation (decompose into multiple Graphics objects) is mentioned in the unknowns section but should be elevated to a documented pattern in the API Surface section.

#### Disposal scope integration is correct

`effect()` calls `_tryOnCleanup(dispose)` at line 222 of `signal.ts`. This means effects created inside a `pushScope()`/`popScope()` block automatically register their disposal with the parent scope. The design's `jsxCanvas()` sketch creates effects inside a scope and registers cleanup with `onCleanup()` — both are correct paths.

**One concern:** The sketch shows both explicit `cleanups.push(effect(...))` AND `onCleanup(() => cleanups.forEach(fn => fn()))`. If `jsxCanvas` runs inside a disposal scope (which it does — component calls use `runInScope`), the effects will be registered TWICE: once via `_tryOnCleanup` inside `effect()` itself, and once via the explicit `onCleanup` at the end. This will cause double-disposal.

**Recommendation:** Choose one strategy. Either:
- (A) Rely on `_tryOnCleanup` auto-registration (effects dispose when their parent scope disposes), and only use the explicit cleanup for non-effect resources (event listeners, `displayObject.destroy()`). This is the pattern `__list` and `__conditional` use.
- (B) Run effect creation in an `untrack`-like wrapper that suppresses auto-registration, and manage all cleanup explicitly.

Option (A) is simpler and consistent with the existing codebase. The implementation sketch should be updated to reflect this.

#### Context scope capture in effects

`EffectImpl` captures `getContextScope()` at creation time (line 168 of `signal.ts`) and restores it on re-run (line 187). This means if `jsxCanvas` creates effects inside a `CanvasRenderContext.Provider` scope, those effects will correctly have access to the canvas context on re-run. This is important for correctness and it works automatically. Good.

---

### Risks and Concerns

#### RISK 1: The `jsxImportSource` switching model is incomplete (Medium severity)

The design proposes that canvas component files use `@vertz/ui-canvas/jsx-runtime` as their `jsxImportSource`. This means **the developer must configure their TypeScript/build tool to use a different JSX factory for canvas files**.

This creates a problem: a single file like the `DesignTool` example in the design doc contains BOTH DOM JSX (`<div>`, `<aside>`, `<h2>`, `<input>`) and Canvas JSX (`<Container>`, `<Graphics>`, `<SelectableRect>`). A file can only have one `jsxImportSource`. You cannot have `<div>` resolve to the DOM factory and `<Graphics>` resolve to the canvas factory in the same file.

The design's Unknown #1 proposes a context-based switch, but the `jsxImportSource` mechanism is a compile-time decision, not a runtime decision. The TypeScript compiler transforms ALL JSX in a file through the same factory.

**Possible solutions:**
1. **Unified JSX factory with runtime dispatch.** A single `jsx()` function that checks whether the tag is a known canvas intrinsic (from a registry) and dispatches to `jsxCanvas()` or `jsxDOM()` accordingly. The design rejected this ("adds overhead to all DOM rendering"), but the overhead is a single Map lookup per element — negligible.
2. **`<CanvasLayer>` renders children via explicit call, not JSX.** Children of `<CanvasLayer>` are passed as a render function: `<CanvasLayer render={() => <Container>...</Container>} />`, where the render function's file uses the canvas `jsxImportSource`. This is clunky.
3. **Compiler detects `<CanvasLayer>` and switches factory for its children subtree.** This is the compiler plugin approach the design explicitly deferred.

Solution 1 is the pragmatic answer for Phase 2. The overhead concern is not justified — a `Set.has()` check for ~4 intrinsic names adds nanoseconds per element. The unified factory approach also means developers never have to think about `jsxImportSource` configuration.

**This must be resolved in the design before implementation begins.** The current proposal has a compile-time/runtime mismatch that will block developers from writing hybrid DOM/Canvas files.

#### RISK 2: Dynamic list reconciliation lacks detail (Low-Medium severity)

The design mentions `canvasList()` as analogous to `__list`, but the implementation details are thin. The DOM `__list` implementation (which I have read) relies on DOM-specific operations:
- `container.childNodes[i]` for position comparison
- `container.insertBefore()` for reordering
- `node.parentNode?.removeChild(node)` for removal

The canvas equivalent needs PixiJS-specific operations:
- `container.getChildAt(i)` for position comparison
- `container.addChildAt(child, index)` for insertion at position
- `container.removeChild(child)` for removal
- `container.setChildIndex(child, newIndex)` for reordering

These are semantically equivalent, but the implementation will be a separate code path. The design should explicitly state that `canvasList` is a **new implementation** that mirrors `__list`'s algorithm but uses PixiJS Container methods, not a wrapper around `__list`.

Additionally, the keyed list example uses `() => nodes.value.map(...)` which returns an array from a reactive accessor. The runtime needs to:
1. Detect that a child is an accessor function (not a static array)
2. Wrap it in an `effect()`
3. On re-run, diff the key list against the previous render
4. Add/remove/reorder PixiJS display objects

This is not trivial and deserves its own focused sub-section in the design, particularly around **disposal of removed items**. When an item is removed from the list, its disposal scope must run (to clean up effects and event listeners on that display object). The DOM `__list` handles this via `scopeMap` — the canvas version needs the same pattern.

#### RISK 3: Graphics redraw is a real concern for Phase 2c interactions (Medium severity)

The design correctly identifies `clear() + redraw` as an unknown, but I want to emphasize: this is not just about raw path count. When a `Graphics` object is used for interactive selection handles (the `SelectableRect` pattern in the design tool example), the draw callback will likely read `selected` state:

```typescript
draw={(g) => {
  const sel = unwrap(props.selected);
  g.rect(0, 0, w, h);
  g.fill(sel ? 0x4ecdc4 : 0x333333);
  if (sel) {
    g.rect(-4, -4, 8, 8); // handle
    g.fill(0xffffff);
  }
}}
```

Every time selection changes, this triggers a full redraw. If you have 100 nodes and selection changes from node A to node B, TWO Graphics objects redraw (A deselects, B selects). This is fine. But if a user drags a node and `draw` reads position signals, the redraw happens on every `pointermove` event — potentially 60+ times per second per dragged object.

The POC should specifically test: **reactive drag with Graphics redraw at 60fps**. Not just "100 paths updating one value per frame" but "a Graphics object with 10+ draw commands redrawn every frame while the user drags."

#### RISK 4: Nested `<CanvasLayer>` behavior is undefined (Low severity)

The design does not address what happens if a developer writes:

```tsx
<CanvasLayer>
  <Container>
    <CanvasLayer> {/* Nested! */}
      <Graphics ... />
    </CanvasLayer>
  </Container>
</CanvasLayer>
```

This should either be explicitly forbidden (throw an error) or explicitly supported with defined behavior. Given that `<CanvasLayer>` creates a DOM `<div>` with a `<canvas>` inside it, nesting would mean creating a DOM element inside a canvas scene graph, which is nonsensical.

**Recommendation:** Add a runtime check in `<CanvasLayer>` that throws if it detects it is already inside a canvas rendering context. Add this to the Non-Goals section.

#### RISK 5: `unwrap` utility is load-bearing but underspecified (Low severity)

The `unwrap` utility appears in the `Circle` component example:
```typescript
const r = unwrap(props.radius);
```

This function resolves `T | (() => T)` to `T`. When called inside an `effect`, if the value is an accessor, calling it tracks the signal dependency. This is correct and important.

However, `unwrap` is not defined anywhere in the current codebase. It needs to be:
1. Exported from `@vertz/ui-canvas`
2. Properly typed so `unwrap<T>(value: T | (() => T)): T` preserves the type
3. Documented as the way to read reactive-or-static props inside draw callbacks

This is a minor gap but should be called out in the design doc since it is part of the public API surface.

---

### Recommendations

1. **Resolve the JSX factory dispatch mechanism (RISK 1) before implementation.** This is a blocking issue. I recommend the unified factory with runtime dispatch approach. Update the design doc to specify this.

2. **Fix the double-disposal issue in the `jsxCanvas` sketch.** Remove the explicit `cleanups.push(effect(...))` pattern and rely on `_tryOnCleanup` auto-registration. Only manually track non-effect cleanups (event listeners, `displayObject.destroy()`).

3. **Expand the `canvasList` design.** Add a subsection under "Dynamic Children" that shows the PixiJS-specific reconciliation algorithm. It should mirror `__list`'s structure: `nodeMap`, `scopeMap`, key-based diff, disposal of removed items.

4. **Add `unwrap` to the API Surface section.** Define its type signature, export location, and note that it is essential for custom canvas components that accept `T | (() => T)` props.

5. **Explicitly forbid nested `<CanvasLayer>`** in the Non-Goals section and add a runtime check that throws.

6. **Add a drag-while-redrawing benchmark** to the Graphics POC criteria. The current POC spec ("100+ paths, update one reactive value per frame") does not cover the interactive drag use case which is the primary design tool scenario.

7. **Document the "all deps trigger full redraw" behavior of `draw` callbacks.** This is not a bug — it is inherent to the fine-grained model — but developers need to know. Include a "Performance Tips" subsection that recommends decomposing complex Graphics into multiple objects when different parts react to different signals.

---

### Type Flow Map Assessment

The type flow map is adequate but could be stronger. The per-element prop interfaces (`GraphicsProps`, `SpriteProps`, etc.) are listed, and the `.test-d.ts` requirement is stated. However, the map does not trace the flow from **user-defined component props** through `jsxCanvas` to the display object.

Specifically: when a user writes `<Circle x={() => pos.value}>`, the `Circle` function receives `CircleProps` where `x: number | (() => number)`. Inside the component, the user passes `props.x` to `<Graphics x={props.x}>`. The canvas factory then needs to handle both `number` and `() => number` for the same prop. This type widening from the user component to the intrinsic element is correct, but the `.test-d.ts` suite should include a test that verifies passing `() => number` through a custom component to an intrinsic element compiles correctly.

---

### Consistency Check

The design is consistent with Vertz's existing patterns:

- **Signal tracking:** Uses `effect()` identically to `__attr()` and `__text()` in the DOM renderer.
- **Disposal:** Uses `pushScope()`/`popScope()`/`onCleanup()` identically to `__list` and `__conditional`.
- **Context:** Uses `createContext` identically to the DOM context system.
- **Component model:** Function components receiving props, same as DOM.
- **Events:** `on*` prop convention, same as DOM JSX runtime.

The only divergence is the JSX factory dispatch (RISK 1), which needs resolution.

---

### Verdict

**Approve with comments.** The architecture is sound and the signal integration is correct. The no-reconciliation decision is the right call. However, RISK 1 (JSX factory dispatch for hybrid files) is a blocking issue that must be resolved in the design before implementation begins. The remaining items (double-disposal, canvasList detail, unwrap, nested CanvasLayer, drag benchmark) are non-blocking but should be addressed before Phase 2a implementation starts.

After RISK 1 is resolved and the sketch is updated, this design is ready for implementation.

PR #489, opened 2026-02-21
feat(ui,entity-todo): form() attrs() returns onSubmit + full-stack demo

viniciusdacal:
## Summary

- **`form().attrs()` now returns `onSubmit`** — accepts optional `SubmitCallbacks` and returns `{ action, method, onSubmit }` for declarative JSX form wiring. No more manual `addEventListener`. Includes `resetOnSuccess` option.
- **`__attr()` handles boolean values** — `true` → `setAttribute(name, '')`, `false` → `removeAttribute(name)`. Fixes SSR serialization crash where `disabled={false}` stored a boolean in attrs, causing `escapeAttr()` to crash on `.replace()`.
- **FetchClient convenience methods** — `get()`, `post()`, `patch()`, `put()`, `delete()` wrappers on `FetchClient`.
- **CLI codegen command** wired to action handler.
- **Entity-todo upgraded to full-stack demo** — UI components (TodoForm, TodoItem, TodoListPage), SSR tests, mock data, Vite config, test infrastructure (compiler plugin + happy-dom setup).

Closes #485

## Test plan

- [x] `@vertz/ui` — 877 tests pass (form unit tests, type tests, __attr boolean tests, integration tests)
- [x] `entity-todo` — 22 tests pass (API + TodoForm + TodoList + SSR)
- [x] `@vertz/fetch` — convenience method tests pass
- [x] Full monorepo quality gates: lint, typecheck, test (60/60 turbo tasks green)
- [ ] Manual: `bun run dev` in entity-todo — verify CRUD works end-to-end

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #488, opened 2026-02-21
feat(ui): eliminate manual effect() bridge for external reactive state


viniciusdacal:
## Problem

Every component that uses `form()` or `query()` must manually bridge external signals into local `let` variables via `effect()` for the compiler's reactivity system to track them:

```tsx
// This boilerplate is repeated in every form/query component
let titleError = '';
let isSubmitting = false;
let submitLabel = 'Add Todo';

effect(() => {
  titleError = todoForm.error('title') ?? '';
  isSubmitting = todoForm.submitting.value;
  submitLabel = isSubmitting ? 'Adding...' : 'Add Todo';
});
```

The root cause: the compiler transforms `let x` → `signal()` and `{x}` → `__child(() => x.value)`, but it can't see inside external function calls like `todoForm.error('title')` or `todoForm.submitting.value`. These external signals need manual bridging to become reactive in the template.

This pattern appears in:
- `TodoForm` (entity-todo) — `form().error()`, `form().submitting`
- `TodoListPage` (entity-todo) — `query().data`, `query().loading`, `query().error`
- `TaskForm` (task-manager) — `form().error()`, `form().submitting`
- `TaskListPage` (task-manager) — `query().data`, `query().loading`

## Impact

- **DX friction**: Every component using framework APIs repeats this boilerplate
- **Error-prone**: Forgetting a binding means stale UI
- **Violates MANIFESTO.md**: "production-ready by default" — the developer shouldn't need to wire plumbing

### 1. Compiler-level recognition
Teach the compiler to recognize known framework APIs (`form().error()`, `form().submitting`, `query().data`) and auto-generate the effect bridge.

### 2. API-level pre-wrapped bindings
`form()` returns pre-wrapped local bindings:
```tsx
const { titleError, isSubmitting } = todoForm.bind({
  titleError: () => todoForm.error('title') ?? '',
  isSubmitting: () => todoForm.submitting.value,
});
```

### 3. Reactive directive primitive
A `bind()` or `use()` that tells the compiler "this expression is reactive":
```tsx
let titleError = bind(() => todoForm.error('title'));
```

### 4. Compiler auto-detection of `.value` access
The compiler could detect `signal.value` access patterns in expressions and auto-wrap them in reactive tracking, even for non-`let` sources.

## Requirements

- Zero manual `effect()` calls needed for standard form/query state
- The solution should work for any external reactive source, not just `form()` and `query()`
- Must be type-safe
[...]

PR #491, opened 2026-02-21
fix(ui,fetch,ui-server): address PR #489 review findings

viniciusdacal:
## Summary

Addresses the actionable review findings from PR #489 (DX, Technical, and Process reviews):

- **escapeAttr safety net** — Add defensive `String()` coercion in `escapeAttr()` so manually constructed VNodes with non-string attrs don't crash the SSR serializer
- **Missing form tests** — Add `resetOnSuccess` when SDK throws (verify form NOT reset), `attrs()` called multiple times produces independent handlers, extract shared `createSubmitEvent()` helper to reduce test duplication
- **JSDoc improvement** — Document that `resetOnSuccess` only works with DOM Events (not raw FormData)
- **FetchClient params fix** — Remove incorrect `normalizeOptions()` that mapped `params` → `query`. Params are **path parameters** (`:brandId`), handled by codegen's template literals at the SDK layer, not query string parameters
- **Design doc reconciliation** — Update Section 3C to honestly reflect that `effect()` bridges are still needed for external reactive state (tracked in #488), replacing the "zero effect()" claim

## Test plan

- [x] `escapeAttr` test: non-string attribute values are coerced to strings (11/11 pass)
- [x] Form tests: 22/22 pass including 3 new tests
- [x] Form integration tests: 38/38 pass
- [x] ui-server tests: 126/126 pass
- [x] entity-todo tests: 22/22 pass (API + UI + SSR)
- [x] FetchClient tests: 34/35 pass (1 pre-existing `toHaveBeenCalledBefore` failure)
- [x] Typecheck clean
- [x] All 60 pre-push quality gates pass

### DIFF
--- a/.changeset/ui-server-escape-attr-coercion.md
+++ b/.changeset/ui-server-escape-attr-coercion.md
@@ -1,0 +1,5 @@
+---
+'@vertz/ui-server': patch
+---
+
+escapeAttr() now defensively coerces non-string attribute values to strings instead of crashing.


viniciusdacal:
## PR #491 Technical Review

### 1. `escapeAttr` defensive coercion

The implementation:
```ts
const str = typeof value === 'string' ? value : String(value);
```

**Verdict: Acceptable as a safety net, but has a known rough edge.**

The `typeof` check is cheap and the `String()` fallback handles the realistic cases well: numbers (`42` -> `"42"`), booleans (`true` -> `"true"`). These are the most likely runtime type violations when VNodes are constructed manually.

However, `String()` produces poor output for two cases that could appear if a VNode is constructed carelessly:
- `null` -> `"null"`, `undefined` -> `"undefined"` -- these would serialize as `data-foo="null"` which is valid HTML but almost certainly a bug. A guard that skips the attribute entirely (or throws) would be safer.
- `{a:1}` -> `"[object Object]"` -- renders meaningless HTML.

Since the `VNode.attrs` type is `Record<string, string>`, these values would require a deliberate `as unknown as string` cast to reach this code, so the risk is low. The current approach is a reasonable tradeoff: it prevents a crash without over-engineering. If you want to harden this further, consider adding a `null`/`undefined` skip:

```ts
if (value == null) return '';  // or skip the attribute entirely in serializeAttrs
```

Not blocking, but worth a follow-up issue if you want defense-in-depth against `"null"` attributes in SSR output.

---

### 2. Form test correctness

#### `resetOnSuccess` when SDK throws (line 203-215)

**Verdict: Correct.** The test:
1. Configures `mockRejectedValue` on the handler -- the SDK call will throw.
2. Passes `resetOnSuccess: true` along with a no-op `onError` callback (preventing an unhandled rejection).
3. Asserts `mockReset` was NOT called.

This exercises the correct code path: `createSubmitHandler` catches the error in its `try/catch` (line 117-124 of `form.ts`), calls `onError`, and returns *before* reaching the `resetOnSuccess` check at line 127. The test correctly proves reset is gated on success, not just on the option being set. Good coverage.

#### `attrs()` independence (line 146-170)

**Verdict: Correct.** The test creates two `attrs()` calls with different `onSuccess` callbacks, invokes only `attrsA.onSubmit`, and asserts `onSuccessB` was never called. This proves each `attrs()` call creates an independent closure via `createSubmitHandler`, not a shared mutable callback slot. The test is precise and the assertion is sufficient.

#### `createSubmitEvent` helper (line 45-72)

**Verdict: Correct and well-structured.** The helper:
1. Creates a real `Event('submit')` with `cancelable: true`.
2. Patches `event.target` to a mock form element with `reset()`.
3. Monkey-patches `globalThis.FormData` so `new FormData(formElement)` returns the desired entries.
4. Returns a `cleanup` function to restore the original `FormData`.
5. Every caller wraps the test in `try/finally` to guarantee cleanup.

The `extends OriginalFormData` approach is clean -- it preserves the prototype chain so `instanceof FormData` checks still work in the production code. The `_formElement` parameter is correctly ignored (the mock doesn't need a real DOM element).

One minor observation: if a test throws unexpectedly before `cleanup()` is called, the patched `FormData` would leak into subsequent tests. The `try/finally` blocks handle this correctly, so this is not a problem in practice. Could also be done with `beforeEach`/`afterEach`, but the current approach is explicit and easier to follow.

---

### 3. FetchClient `params` removal

**Verdict: Correct. This is a clean removal of dead code.**

Evidence:
- **`RequestOptions` (types.ts)** does not include a `params` field. The removed `normalizeOptions` was casting to `RequestOptions & { params?: Record<string, unknown> }` -- an ad-hoc extension that was never part of the public type contract.
- **Codegen (`emit-client.ts` line 69)** handles path parameters via template literal interpolation: `/:brandId` becomes `` `${input.params.brandId}` ``. The `params` object is consumed *at the SDK layer* before any FetchClient method is called. The generated SDK methods call `client.request('GET', interpolatedPath, { query: input?.query })` -- params are already baked into the path string.
- **No remaining references** to `normalizeOptions` or `params` in the `packages/fetch` directory.
- The two removed tests (`get() maps params to query` and `get() preserves query when both params and query are provided`) tested the removed behavior with `as any` casts, confirming this was never type-safe.

The changeset update in `fetch-convenience-methods.md` correctly documents this removal.

---

### 4. Design doc changes (Section 3C)

**Verdict: Accurate and honest.**

The previous claim of "zero `effect()`" was incorrect -- the compiler cannot track external signal reads from `form()` APIs. The updated section:
- Correctly describes the limitation: the compiler only transforms local `let` variables into signals.
- Shows the required `effect()` bridge pattern with a concrete code example.
- References #488 as the tracking issue. I verified #488 exists, is open, and its description matches -- it tracks eliminating the manual `effect()` bridge for external reactive state.
- The Phase 6/7 step updates (lines 468-493) consistently replace "no `effect()`" claims with "via `effect()` bridge" language.

The "Reduced `effect()`" phrasing in the section heading and properties list is accurate -- `addEventListener` and manual DOM wiring are genuinely eliminated, only the signal bridge remains.

---

### 5. JSDoc on `resetOnSuccess`

**Verdict: Clear and sufficient.**

```ts
/**
 * When true, the form element is reset after a successful submission.
 * Only takes effect when the handler receives a DOM Event (so `formElement`
 * can be extracted from `event.target`). Has no effect with raw FormData.
 */
resetOnSuccess?: boolean;
```

The doc explains both the positive case (DOM Event -> `event.target` -> `formElement.reset()`) and the no-op case (raw FormData). The parenthetical explains *why* it requires a DOM Event, which helps developers understand the mechanism rather than just memorizing the rule. This aligns with how `createSubmitHandler` works: the `formElement` variable is only set in the `instanceof Event` branch (line 90-93 of `form.ts`), and the reset guard checks `callbacks?.resetOnSuccess && formElement` (line 127).

---

### Summary

This is a clean follow-up PR. The changes are well-scoped, the tests are correct, the `normalizeOptions` removal is justified by the codegen architecture, and the design doc updates are honest about current limitations. No blocking issues.

Minor suggestion for a follow-up: consider whether `escapeAttr` should skip or warn on `null`/`undefined` values rather than serializing them as literal `"null"`/`"undefined"` strings. Low priority since the type system guards against this.

viniciusdacal:
## DX & Process Review -- PR #491

**Reviewer role:** DX and process compliance  
**PR under review:** #491 (address PR #489 review findings)  
**Original review:** PR #489 (DX, Technical, and Process reviews)

---

### 1. Process Compliance

#### 1a. Commit Message Format (`commits.md`)

**Commit headline:** `fix(ui,fetch,ui-server): address PR #489 review findings [#489]`

- Format `<type>(<scope>): <description> [#<issue>]` -- **correct**.
- Issue reference `[#489]` at end of subject line -- **correct**.
- Multi-scope `(ui,fetch,ui-server)` accurately reflects the three packages touched -- **acceptable**.
- `Co-Authored-By` trailer present -- **correct**.
- Commit body lists each change as a bullet -- **good practice for a multi-concern fix commit**.

**Verdict: Pass.**

#### 1b. TDD Compliance (`tdd.md`)

This PR addresses review findings rather than implementing new features, so strict Red-Green-Refactor sequencing is less critical. However, the rule states: "Never write implementation code without a failing test."

Looking at the changes:
- **`escapeAttr` coercion** -- test added (`'coerces non-string attribute values to strings'`) alongside the implementation change. The test is present, verifying the defensive behavior.
- **`resetOnSuccess` on SDK error** -- new test added (`'onSubmit with resetOnSuccess does NOT reset the form when SDK throws'`). This tests a code path that already existed in the implementation; the test is a regression guard, not driving new behavior. Appropriate for a review-findings PR.
- **`attrs()` independence** -- new test added (`'returns independent handlers when called multiple times'`). Again, tests an existing behavior that was undercovered.
- **FetchClient params removal** -- tests that used `as any` to call the incorrect `params` API are removed alongside the `normalizeOptions` method. Test removal + implementation removal is coherent.
- **JSDoc and design doc changes** -- no behavioral changes, so no tests required.

**Verdict: Pass.** All behavioral changes have corresponding tests. The test-first ordering cannot be verified from a single commit, but for a review-findings PR this is acceptable.

#### 1c. Changeset (`semver.md`)

The diff modifies the existing `.changeset/fetch-convenience-methods.md`:
- Listed package: `@vertz/fetch` with `patch` -- **correct per `semver.md`**.
- Description updated to include: "Removed incorrect params-to-query mapping."

**Issue: Missing changesets for `@vertz/ui` and `@vertz/ui-server`.**

This PR changes:
- `packages/ui/src/form/form.ts` -- JSDoc update to `resetOnSuccess`. This is documentation-only, not a behavioral change. Arguably no changeset needed.
- `packages/ui-server/src/html-serializer.ts` -- `escapeAttr` now coerces non-string values. This IS a behavioral change (previously it would crash; now it coerces gracefully). It should have its own changeset entry.

The `@vertz/ui-server` change is a defensive safety net in a function that was already supposed to receive strings (per its TypeScript signature). It could be argued both ways: it is a bug fix (crash prevention) or an internal-only safety net that does not change the public API contract. However, per `definition-of-done.md`, any behavior change to a published package needs a changeset.

**Verdict: Minor gap.** `@vertz/ui-server` should have a `patch` changeset for the `escapeAttr` coercion fix. The `@vertz/ui` JSDoc change is documentation-only and arguably does not require one.

#### 1d. No `@ts-ignore` or `as any`

- The diff **removes** two `as any` casts from the deleted params-to-query tests. No new `as any` introduced.
- One `as unknown as string` exists in the new `html-serializer.test.ts` test -- this is intentional: the test deliberately passes a non-string value to verify runtime coercion. This is a legitimate test cast pattern per `biome.md` (the `no-double-cast` rule notes double casts are "legitimate in test mocks").
- No `@ts-ignore` anywhere.

**Verdict: Pass.**

---

### 2. DX Impact -- JSDoc Improvements

The `resetOnSuccess` JSDoc now reads:

```ts
/**
 * When true, the form element is reset after a successful submission.
 * Only takes effect when the handler receives a DOM Event (so `formElement`
 * can be extracted from `event.target`). Has no effect with raw FormData.
 */
resetOnSuccess?: boolean;
```

**Assessment:** This is exactly what the #489 DX review requested. A developer reading this JSDoc would understand:
1. What `resetOnSuccess` does (resets form after success).
2. When it works (DOM Event path only).
3. When it does NOT work (raw FormData path).
4. Why (the handler needs `event.target` to find the form element).

The parenthetical `(so formElement can be extracted from event.target)` gives just enough implementation context for the "why" without over-explaining. A developer using `attrs()` (which always produces an `onSubmit` that receives events) would understand their form will reset. A developer using `handleSubmit()` with raw FormData would understand it will not.

**Verdict: Clear and actionable. Pass.**

---

### 3. Test Readability -- `createSubmitEvent` Helper

The `createSubmitEvent` helper (lines 45-72 of `form.test.ts`) replaces 15+ lines of duplicated mock setup across three tests. Evaluating its impact:

**Improvements:**
- Each test using it is now 6-8 lines of assertion logic instead of 20+ lines of setup boilerplate.
- The helper name `createSubmitEvent` is descriptive -- a new developer can infer it creates a fake submit event.
- The JSDoc explains what it does and mentions the `FormData` patching.
- The `cleanup` return value makes the `try/finally` pattern explicit -- no risk of leaked global state.
- The `entries` parameter (a simple `Record<string, string>`) is intuitive -- you specify what form data you want.

**Potential concerns:**
- The helper patches `globalThis.FormData`, which is a global side effect. The `try/finally` + `cleanup` pattern mitigates this, but a new developer might not understand WHY the helper needs to patch FormData. A brief inline comment like `// Patch FormData so new FormData(event.target) returns our entries` would help, but the existing JSDoc ("patches globalThis.FormData so `new FormData(target)` returns the provided entries") already covers this.
- The helper returns `{ event, mockReset, cleanup }` -- three things. A new developer needs to understand what `mockReset` is (the spy on `form.reset()`). The name is self-documenting enough.

**Test examples after the refactor:**

```ts
it('onSubmit with resetOnSuccess resets the form element on success', async () => {
  // ... 3 lines of setup ...
  const { event, mockReset, cleanup } = createSubmitEvent({ name: 'Alice' });
  try {
    await f.handleSubmit({ resetOnSuccess: true })(event);
    expect(mockReset).toHaveBeenCalled();
  } finally {
    cleanup();
  }
});
```

This is significantly more readable than the original 25-line version. Each test's intent is immediately clear from the `expect` assertion.

**Verdict: Clear improvement. Pass.**

---

### 4. Design Doc Honesty -- Section 3C

The updated Section 3C header changed from:
> "Declarative form state -- zero `effect()`"

to:
> "Declarative form state -- reduced `effect()` (with known limitation)"

**Assessment of the explanation:**

The section now:
1. Shows the *ideal* JSX usage (what it would look like without the limitation).
2. Explains the *limitation* clearly: "The vertz compiler only transforms local `let` variables into reactive signals."
3. Shows the *workaround* with a concrete code example (effect bridges).
4. Links to the tracking issue (#488).
5. States the *planned fix* and when it will eliminate the workaround.

The implementation plan sections (Phases 5 and 6) are also updated to reflect `effect()` bridges instead of "zero effect()". The summary properties table changed from "Zero `effect()`" to "Reduced `effect()` -- ... `effect()` bridges still needed for external reactive state (see Section 3C, tracked in #488)."

**Would a developer understand the limitation?** Yes. The explanation is straightforward:
- Compiler tracks `let` variables -> transforms them to signals.
- External signals (from `form()`) are not `let` variables -> compiler cannot track them.
- Bridge: assign external signal values to local `let` variables inside `effect()`.
- This is temporary, tracked in #488.

The only potential confusion is the first code block showing the ideal JSX without context that "this doesn't work yet." However, the **Current limitation** bold text immediately after disambiguates this. A developer reading linearly would understand.

**Verdict: Honest and accurate. The "effect() bridge" is well-explained. Pass.**

---

### 5. Scope

Files changed:
1. `.changeset/fetch-convenience-methods.md` -- updated changeset description
2. `packages/fetch/src/client.test.ts` -- removed incorrect params-to-query tests
3. `packages/fetch/src/client.ts` -- removed `normalizeOptions()`, simplified convenience methods
4. `packages/ui-server/src/__tests__/html-serializer.test.ts` -- added `escapeAttr` coercion test
5. `packages/ui-server/src/html-serializer.ts` -- added defensive `String()` coercion
6. `packages/ui/src/form/__tests__/form.test.ts` -- added 3 tests, extracted helper, reordered
7. `packages/ui/src/form/form.ts` -- JSDoc improvement on `resetOnSuccess`
8. `plans/form-attrs-api-improvement.md` -- reconciled Section 3C and implementation plan steps

Every change traces back to a specific actionable item from the #489 review. No unrelated changes are included. The original #489 review flagged scope issues (CLI wiring, codegen changes, etc.) -- none of those are in this PR.

**Verdict: Appropriately scoped. Pass.**

---

### 6. Cross-Reference with PR #489 Review Findings

Mapping each actionable item from the three #489 reviews to this PR:

#### From the DX Review:

| # | Finding | Status | Notes |
|---|---------|--------|-------|
| 1 | Reconcile Section 3C "zero effect()" with actual effect bridge | **Addressed** | Section 3C rewritten with limitation, workaround, and #488 tracking link |
| 2 | JSDoc on `resetOnSuccess` -- document DOM-event-only limitation | **Addressed** | JSDoc updated with clear explanation |
| 3 | Extract shared `createSubmitEvent` helper for resetOnSuccess tests | **Addressed** | Helper created, three tests refactored to use it |
| 4 | Flat API (`form()` returns `{ action, method, ... }` directly) | Not addressed | This was a "consider" recommendation, not a required fix. Appropriate to defer. |

#### From the Technical Review:

| # | Finding | Status | Notes |
|---|---------|--------|-------|
| 1 | Defensive `String()` coercion in `escapeAttr` | **Addressed** | `escapeAttr` now coerces with `typeof` check + `String()` |
| 2 | Test `resetOnSuccess` when SDK throws | **Addressed** | New test: `'does NOT reset the form when SDK throws'` |
| 3 | Test `attrs()` called multiple times -- independent handlers | **Addressed** | New test: `'returns independent handlers when called multiple times'` |
| 4 | `params` uses `as any` -- fix the incorrect mapping | **Addressed** | `normalizeOptions()` removed entirely; `as any` tests removed; changeset updated |
| 5 | Missing negative type test for `attrs()` | **Not addressed** | The `@ts-expect-error` test for wrong callback shape was not added |
| 6 | Concurrent submission guard/documentation | Not addressed | This was a "worth documenting" suggestion, not a required fix |
| 7 | `onSubmit` with `event.target === null` edge case | Not addressed | This was flagged as "low risk" |

#### From the Process Review:

| # | Finding | Status | Notes |
|---|---------|--------|-------|
| 1 | Missing changeset | **Partially addressed** | `@vertz/fetch` changeset updated; `@vertz/ui-server` still missing (see 1c above) |
| 2 | Design deviation (codegen/CLI out of scope) | **Addressed** | Those changes are not in this PR -- scope is correct |
| 3 | PR scope too broad | **Addressed** | This PR is tightly scoped to review findings only |
| 4 | Missing negative type tests | **Not addressed** | Same as Technical #5 above |

---

### Summary

| Area | Verdict |
|------|---------|
| Commit message format | **Pass** |
| TDD compliance | **Pass** (acceptable for review-findings PR) |
| Changeset | **Minor gap** -- missing `@vertz/ui-server` patch changeset |
| No `@ts-ignore`/`as any` | **Pass** |
| JSDoc clarity | **Pass** |
| Test readability | **Pass** -- `createSubmitEvent` is a clear improvement |
| Design doc honesty | **Pass** -- Section 3C is now accurate |
| Scope | **Pass** -- tightly scoped to #489 findings |
| Review findings coverage | **Good** -- all priority items addressed; 2 minor items deferred |

### Items to Address

**Non-blocking (should be tracked):**

1. **Missing `@vertz/ui-server` changeset.** The `escapeAttr` coercion is a behavioral change (crash -> graceful coercion). Add a `patch` changeset for `@vertz/ui-server`.

2. **Missing negative type test for `attrs()`.** The #489 technical review specifically called out: "no `@ts-expect-error` test verifying that `attrs()` rejects an invalid callback shape." This was listed as an actionable item but was not addressed in this PR. Consider adding in a follow-up.

### Recommendation

**Approve.** This PR is well-scoped, addresses the priority findings from #489, and improves both test readability and documentation honesty. The two non-blocking items above can be addressed in follow-up work. The `@vertz/ui-server` changeset gap is the more pressing of the two -- it should be added before merge if possible, but it is not blocking given the change is a defensive safety net rather than a public API contract change.

Issue #561, opened 2026-02-22
feat(ui): Add RouterProvider to bridge router state and Outlet rendering


viniciusdacal:
## Problem

The router has all the building blocks for declarative route rendering — `createRouter()` (reactive `current` signal), `createOutlet()` (declarative Outlet component), and `OutletContext` — but there's **no built-in component that wires them together**.

Today, the app shell must manually watch `router.current` and swap DOM content imperatively:

```tsx
// Current: imperative DOM swapping in app.tsx
function updateContent(node: Node) {
  main.innerHTML = '';
  main.appendChild(node);
}

watch(
  () => appRouter.current.value,
  (match) => {
    if (!match) {
      updateContent(<div>Page not found</div>);
      return;
    }
    const component = match.route.component();
    if (component instanceof Promise) {
      component.then((mod) => {
        updateContent((mod as { default: () => Node }).default());
      });
    } else {
      updateContent(component);
    }
  },
);
```

This is exactly the kind of imperative DOM manipulation we've been eliminating from examples. It also forces every app to rewrite the same boilerplate (match handling, async component resolution, not-found fallback).

## Proposed Solution

Add a `RouterProvider` component that:

1. Takes the router instance + outlet context
2. Watches `router.current` reactively
3. Populates the `OutletContext` with the matched route's component
4. Handles not-found (no match) with a configurable fallback
5. Handles async/lazy components (`Promise<{ default: () => Node }>`)
6. Optionally integrates View Transitions API

### Target DX

```tsx
// router.ts — unchanged
export const appRouter = createRouter(routes, initialPath);
export const Outlet = createOutlet(outletContext);

// app.tsx — fully declarative
export function App() {
  return (
    <RouterProvider router={appRouter} context={outletContext}>
      <div class={layoutStyles.classNames.shell}>
        <nav>
          <Link href="/" children="All Tasks" />
          <Link href="/tasks/new" children="Create Task" />
          <Link href="/settings" children="Settings" />
        </nav>
        <main>
          <Outlet />
        </main>
      </div>
    </RouterProvider>
  );
}
```

Zero `watch()`, zero `innerHTML`, zero `appendChild` in app code.

### API Surface

```tsx
interface RouterProviderProps {
  router: Router;
  context: Context<OutletContext>;
  /** Rendered when no route matches. Defaults to a "Page not found" element. */
  notFound?: () => Node;
  /** Enable View Transitions API for page swaps (default: true if available). */
  viewTransitions?: boolean;
[...]

PR #564, opened 2026-02-22
feat(ui): add RouterContext, useRouter(), and RouterView [#561]

viniciusdacal:
## Summary

- **RouterContext + useRouter()**: Context-based router access eliminates `navigate` prop threading from every page component
- **RouterView**: Declarative component that reactively renders the matched route's component, replacing manual `watch()` + DOM swapping in app shells
- **Task-manager example rewritten**: All 4 pages now use `useRouter()` instead of receiving `navigate` as a prop; `app.tsx` uses `RouterView` instead of imperative rendering

## Design Doc

See `plans/router-context-view.md` — approved by josh (DX), pm (scope), nora (technical feasibility), and CTO.

## Key Decisions

- **`useRouter()` over `router()`**: Avoids naming collision (`const router = router()` shadows itself), and establishes naming category distinction — primitives (creators) use no prefix, context accessors (readers) use `use` prefix
- **RouterView wraps component factories in `RouterContext.Provider`**: Ensures `useRouter()` works during component construction for both sync and async components
- **Render generation counter**: Guards against stale async component resolutions during rapid navigation

## New Files

| File | Purpose |
|------|---------|
| `packages/ui/src/router/router-context.ts` | `RouterContext` + `useRouter()` |
| `packages/ui/src/router/router-view.ts` | `RouterView` component |
| `packages/ui/src/router/__tests__/router-context.test.ts` | 3 tests |
| `packages/ui/src/router/__tests__/router-view.test.ts` | 12 tests |
| `plans/router-context-view.md` | Design doc |
| `.changeset/router-context-view.md` | Patch changeset |

## Known Regressions

- **View Transitions temporarily lost**: The previous `app.tsx` used `document.startViewTransition()` for smooth page transitions. RouterView does plain DOM swapping. View Transitions support will be added in a follow-up via a `transition` option on RouterView.

## Test plan

- [x] 15 new unit tests (3 context + 12 view) — all pass
- [x] Type tests for `useRouter()` and `RouterView` — typecheck clean
- [x] Subpath exports test updated — verifies no internal leaks
- [x] `@vertz/ui` typecheck clean
- [x] `@vertz/ui` lint clean (only pre-existing warnings)
- [ ] Manual: `cd examples/task-manager && bun run dev` — all pages render, navigation works, params read correctly

### DIFF
--- a/.changeset/router-context-view.md
+++ b/.changeset/router-context-view.md
@@ -1,0 +1,10 @@
+---
+'@vertz/ui': patch
+---
+
+Add RouterContext, useRouter(), and RouterView for declarative route rendering.
+
+- `RouterContext` + `useRouter()`: Context-based router access eliminates navigate prop threading
+- `RouterView`: Declarative component that reactively renders the matched route, replacing manual watch + DOM swapping
+- Handles sync and async/lazy components with stale resolution guards
+- Task-manager example updated to use the new APIs
--- a/.claude/rules/ui-components.md
+++ b/.claude/rules/ui-components.md
@@ -209,3 +209,60 @@ export function useSettings(): SettingsContextValue {
   return ctx;
 }
 ```
+
+## Router
+
+### `RouterContext` + `useRouter()` for navigation
+
+Pages access the router via `useRouter()` context instead of receiving a `navigate` prop. This eliminates prop threading and keeps page signatures clean.
+
+```tsx
+// App shell — wrap in RouterContext.Provider
+RouterContext.Provider(appRouter, () => {
+  const view = RouterView({ router: appRouter });
+  // ... shell layout uses {view}
+});
+
+// Page component — access router via context
+export function TaskListPage() {
+  const { navigate } = useRouter();
+  // navigate('/tasks/new')
+}
+
+// Page with route params — read from router.current
+export function TaskDetailPage() {
+  const router = useRouter();
+  const taskId = router.current.value?.params.id ?? '';
+  // taskId is read once at construction (untracked), which is correct
+}
+```
+
+### `RouterView` for declarative route rendering
+
+Use `RouterView` instead of manual `watch()` + DOM swapping. It handles sync components, async/lazy components, stale resolution guards, and page cleanup automatically.
+
+```tsx
+// WRONG — manual imperative route rendering
+watch(() => appRouter.current.value, (match) => {
+  main.innerHTML = '';
+  main.appendChild(match.route.component());
+});
+
+// RIGHT — declarative with RouterView
+const view = RouterView({
+  router: appRouter,
+  fallback: () => <div>Page not found</div>,
+});
+```
+
+### Route definitions — no prop threading
+
+Route component factories call page functions without props. Pages get everything they need from `useRouter()`.
+
+```tsx
+// WRONG — threading navigate through every route
+'/': { component: () => TaskListPage({ navigate: (url) => router.navigate(url) }) }
+
+// RIGHT — pages use useRouter() internally
+'/': { component: () => TaskListPage() }
+```
--- a/examples/task-manager/src/app.tsx
+++ b/examples/task-manager/src/app.tsx
@@ -5,11 +5,12 @@
  * - JSX for layout composition
  * - ThemeProvider for theme context (CSS variable scoping)
  * - createContext / useContext for app-wide settings
- * - watch() for reacting to external signals (route changes, theme changes)
+ * - RouterContext + RouterView for declarative route rendering
+ * - watch() for reacting to external signals (theme changes)
  * - Full composition of all @vertz/ui features
  */
 
-import { css, ThemeProvider, watch } from '@vertz/ui';
+import { css, RouterContext, RouterView, ThemeProvider, watch } from '@vertz/ui';
 import { createSettingsValue, SettingsContext } from './lib/settings-context';
 import { appRouter, Link } from './router';
 import { layoutStyles } from './styles/components';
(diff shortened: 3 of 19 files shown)


viniciusdacal:
# Technical Feasibility Review -- nora

**PR:** #564 feat(ui): add RouterContext, useRouter(), and RouterView [#561]
**Reviewer:** nora (Frontend & API Surface Engineer)
**Design doc:** `plans/router-context-view.md`

---

## Verdict: APPROVE

This implementation is correct, minimal, and faithful to the design doc. I traced the context scope, effect lifecycle, and async patterns through the runtime source code. No blocking issues found. Observations below are informational or non-blocking.

---

## 1. Context Scope Correctness

**Sync components -- correct.** In `router-view.ts:27`, the sync path wraps `match.route.component()` inside `RouterContext.Provider(router, () => { ... })`. Looking at `context.ts:32-46`, `Provider` pushes the value onto `ctx._stack` and sets `currentScope` for the duration of `fn()`. So when a sync component factory calls `useRouter()`, it hits the fast path in `useContext` (`ctx._stack.length > 0`, line 62) and gets the router. Correct.

**Async components -- correct, with the right subtlety handled.** The `.then()` callback runs in a microtask, outside the original Provider's synchronous scope. The stack has been unwound, and the `currentScope` set by the outer Provider has been restored to its previous value. The implementation correctly wraps the async resolution in a **fresh** `RouterContext.Provider` call (line 33-36). This means the lazy-loaded component's factory gets the router from the stack. Correct.

**Watch re-runs after Provider returns -- correct.** This was the key concern I verified during design review. When `watch()` is called inside a Provider, `effect()` is created (`lifecycle.ts:36`), and `EffectImpl` captures `getContextScope()` at construction time (`signal.ts:168`). On subsequent re-runs (navigation changes), `_run()` restores the captured context scope via `setContextScope(this._contextScope)` (signal.ts:187). So `useContext()` falls through to the async path (line 67 in context.ts: `currentScope?.has(key)`) and returns the correct value. The test `useRouter works inside watch callback via captured context scope` validates this path explicitly.

**No scope issues found.**

## 2. Race Condition Handling

The `renderGen` counter pattern (router-view.ts:14, incremented at line 17) is the standard generation-counter guard for stale async resolutions. The check at line 32 (`if (gen !== renderGen) return`) ensures a late-resolving Promise does not overwrite a more recent navigation's content.

**Edge cases considered:**

- **Rapid A -> B -> A navigation:** Each navigation increments `renderGen`, so even if the user navigates back to a route they came from, the old async resolution for the first visit is correctly discarded. The new visit starts a fresh generation. Correct.
- **Multiple pending async resolutions:** If routes /a, /b, /c are all async and the user navigates a -> b -> c quickly, each navigation increments the counter. Only the most recent (c) will have `gen === renderGen`. Correct.
- **Sync navigation replacing async:** If the user navigates from an async route to a sync route, `container.innerHTML = ''` (line 18) immediately clears, the sync component is appended, and when the stale async resolves, `gen !== renderGen` discards it. Correct.

**One theoretical edge case (non-blocking):** The `renderGen` counter is a plain `number`. In JavaScript, `Number.MAX_SAFE_INTEGER` is 2^53 - 1. At one navigation per second, overflow would take ~285 million years. Not a real concern, but worth noting for posterity.

**No race condition issues found.**

## 3. Cleanup/Disposal

**Watch inner scope cleanup -- correct.** Looking at `lifecycle.ts:34-52`, `watch()` creates an inner cleanup scope via `pushScope()` on every re-run, and runs `runCleanups(innerCleanups)` at the top of each re-execution (line 38-40). This means when the user navigates from route A to route B, the watch callback re-runs, and any cleanups registered during route A's component construction (inside `onMount(() => { onCleanup(...) })`) are executed.

More precisely: `onMount` (`lifecycle.ts:10-25`) pushes a scope, runs the callback, and forwards captured cleanups to the parent scope via `_tryOnCleanup`. The parent scope here is the watch's inner scope (because `onMount` is called during the watch callback execution). When the watch re-fires on navigation, the old inner scope is cleaned up.

The test `page onCleanup runs when navigating away` validates this exact flow.

**Watch effect disposal on router.dispose() -- partial coverage.** The `watch()` in RouterView registers its effect disposal with the current cleanup scope via `_tryOnCleanup` (lifecycle.ts:56-61). If RouterView is called inside a scope that eventually gets cleaned up (e.g., if the app root is ever disposed), the watch is properly stopped. However, RouterView does not expose its own dispose mechanism, and `router.dispose()` only removes the popstate listener and aborts in-flight loaders -- it does not trigger the watch's disposal. In practice, RouterView lives for the lifetime of the application, so this is not a real issue. The design doc acknowledges this is app-lifetime infrastructure.

**No cleanup issues found.**

## 4. Memory Leaks

**Context scope capture in effect -- properly cleaned up.** `EffectImpl._dispose()` sets `_contextScope = null` (signal.ts:208), which releases the captured `Map` reference. When the watch is eventually disposed, the context scope Map (and everything it references) becomes eligible for GC. Correct.

**Watch inner scope cleanup -- prevents accumulation.** Each re-run of the watch callback calls `runCleanups(innerCleanups)` which empties the array (`cleanups.length = 0`, disposal.ts:77) and runs all registered disposal functions. No accumulation of stale closures across navigations.

**Container innerHTML clearing -- removes DOM references.** Setting `container.innerHTML = ''` on each navigation removes old child nodes from the DOM tree, allowing GC of the old page component's entire subtree.

**Async promise `.then()` closure -- no leak concern.** The `.then()` callback captures `gen`, `renderGen`, `router`, and `container`. `gen` and `renderGen` are primitives. `router` and `container` are app-lifetime objects. The Promise itself is short-lived (resolves once). No leak path here.

**No memory leak issues found.**

## 5. Type Safety

**`RouterContext: Context<Router>` annotation -- correct and required.** With `isolatedDeclarations: true` (confirmed in `packages/ui/tsconfig.json:4`), `createContext<Router>()` cannot have its return type inferred without resolving the function body. The explicit `: Context<Router>` annotation satisfies `isolatedDeclarations`. Both `Context` (exported from index.ts:4) and `Router` (exported from navigate.ts) are importable named types. Correct.

**`useRouter(): Router` return type -- correct.** Explicit annotation, required by `isolatedDeclarations`.

**`RouterView({ router, fallback }: RouterViewProps): HTMLElement` -- correct.** Explicit return type annotation. The actual return value `container` is `HTMLDivElement` which is assignable to `HTMLElement`. The design doc specifies `HTMLElement` as the return type to avoid over-constraining.

**Type test coverage (router.test-d.ts):**
- `useRouter()` returns `Router` -- positive test
- `RouterView` returns `HTMLElement` -- positive test
- `RouterView({})` -- `@ts-expect-error` negative test (missing router)
- `RouterView({ router: 'not-a-router' })` -- `@ts-expect-error` negative test (wrong type)
- `RouterViewProps` structure test -- positive test

These cover the design doc's type test assertions from Section 5. Adequate.

**`mod as { default: () => Node }` cast in router-view.ts:34 -- acceptable.** The `component()` return type is `Node | Promise<{ default: () => Node }>` (from define-routes.ts:20). After the `instanceof Promise` check, TypeScript infers the Promise's resolved type but the `mod` parameter in `.then()` gets the Promise's generic argument. The cast is correct and aligns with the `define-routes.ts` type signature. It could be tightened with a generic helper but this is a minor style preference, not a correctness issue.

**No type safety issues found.**

## 6. Test Coverage

15 tests total (3 context + 12 view). Let me assess coverage against the design doc's acceptance criteria:

| Behavior | Test | Status |
|----------|------|--------|
| Initial sync render | `renders matched route sync component` | Covered |
| Navigation swaps content | `swaps content when router.current changes` | Covered |
| useRouter() inside sync component | `provides RouterContext to sync page components` | Covered |
| useRouter() inside async component | `provides RouterContext to async page components` | Covered |
| useRouter() after navigation | `useRouter works in page after navigation` | Covered |
| Route params via useRouter() | `page component reads params via useRouter()` | Covered |
| useRouter() outside Provider | `useRouter throws when called outside...` | Covered |
| No match, no fallback | `renders empty when no route matches and no fallback` | Covered |
| No match, with fallback | `renders fallback when no route matches` | Covered |
| Async/lazy component | `resolves async/lazy component` | Covered |
| Stale async discard | `discards stale async component on rapid navigation` | Covered |
| onCleanup on navigation | `page onCleanup runs when navigating away` | Covered |
| watch context scope capture | `useRouter works inside watch callback via captured context scope` | Covered |
| Returns HTMLDivElement | `returns an HTMLDivElement` | Covered |
| RouterViewProps type | Type tests in router.test-d.ts | Covered |

**Potentially missing edge cases (non-blocking, suggestions for follow-up):**

1. **Fallback called when navigating TO a non-matching route:** Tests cover initial no-match with fallback, but not navigating from a valid route to a non-matching one. Minor gap -- the implementation handles it (the watch callback fires, match is null, fallback renders), but it is not explicitly tested.

2. **Async component that rejects (Promise rejection):** The `.then()` callback handles successful resolution, but there is no `.catch()`. If the async component's Promise rejects (e.g., network error loading a lazy chunk), the error is unhandled. This is noted in the design doc as a non-goal (Section 3: "Auto-rendering errorComponent" is explicitly out of scope), but a `console.error` or swallowing the rejection would be more defensive. The test suite does not cover this path. Non-blocking -- follows the design doc's stated non-goals.

3. **Router dispose during pending async resolution:** If `router.dispose()` is called while an async component is still resolving, the `.then()` callback will still fire (Promises are not cancellable). The `gen !== renderGen` guard does not protect against this since `renderGen` is not incremented on disposal. The callback would call `container.appendChild(node)` on a potentially orphaned container. Practically harmless (the container is not in the DOM), but worth noting.

## 7. Pre-existing Issues (Non-blocking)

These are not introduced by this PR but are worth noting:

- **`Router.current` is `Signal<RouteMatch | null>`, not `ReadonlySignal`:** The `current` property on `Router` (navigate.ts:23) uses `Signal` (read-write). A consumer could theoretically do `router.current.value = null` and break the router's internal state. This should ideally be `ReadonlySignal<RouteMatch | null>` (same for `loaderData`, `loaderError`, `searchParams`). This is a pre-existing design issue, not introduced by this PR.

- **`currentPath` sync pattern in `examples/task-manager/src/router.ts`:** The PR updates the `currentPath` signal sync to use `watch()` (good), but reads from `window.location.pathname` instead of deriving from the `match` object. This works but is fragile -- it depends on `window.history.pushState` having already been called by `navigate()`. Since `applyNavigation` sets `current.value = match` after `pushState`, this should be fine, but deriving the path from `match` directly would be more robust. The design doc tracks "Link API simplification" as a follow-up.

---

## Summary

The implementation is clean, correct, and minimal. It faithfully implements the design doc. The context scope handling is the most subtle part, and it is handled correctly by leveraging the existing `EffectImpl._contextScope` capture mechanism. The render generation counter is sufficient for stale resolution prevention. Cleanup flows through `watch()`'s inner scope mechanism as designed. Types are correct and `isolatedDeclarations`-compatible.

No blocking issues. The suggestions above (Promise rejection handling, navigate-to-no-match test, ReadonlySignal for Router fields) are all non-blocking and suitable for follow-up tickets.

PR #573, opened 2026-02-22
docs: type-safe router design doc [#572]

viniciusdacal:
## Summary

Design doc for type-safe routing in `@vertz/ui`. Threads route map literal types through the entire router pipeline so that `navigate()`, `useParams()`, and `Link` href are all validated at compile time.

**Requires 3 sign-offs before implementation begins:**
- [ ] DX review (josh)
- [ ] Scope review (pm)
- [ ] Technical feasibility (nora)

## What this covers

- `defineRoutes<const T>()` — preserves literal path keys via phantom generic
- `Router<T>` — typed `navigate(url: RoutePaths<T>)`
- `useParams<TPath>()` — returns `ExtractParams<TPath>` (typed params)
- `createLink<T>()` — typed `href: RoutePaths<T>`
- Full backward compatibility — unparameterized `Router` defaults to `string`
- Type flow map with `.test-d.ts` acceptance criteria for every flow step
- E2E acceptance test

## Key type utilities

- `PathWithParams<T>` — converts `'/tasks/:id'` → `` `/tasks/${string}` ``
- `RoutePaths<T>` — union of all valid URL shapes from a route map
- `TypedRoutes<T>` — branded `CompiledRoute[]` carrying phantom route map type

## Related issues

- Closes #572
- Subsumes #568 (useParams)
- Subsumes #569 (Link API simplification)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #572, opened 2026-02-22
feat(ui): type-safe router — typed navigate, params & links


viniciusdacal:
## Summary

The UI router has a complete `ExtractParams<T>` template literal type utility that correctly extracts typed params from route patterns (e.g., `'/users/:id'` → `{ id: string }`). However, this type information is **lost at 6 key points** before it reaches the developer:

1. `RouteDefinitionMap` uses `[pattern: string]` — erases literal path keys
2. `CompiledRoute` erases `TPath` via `as` cast
3. `Router` is not generic — no type parameter captures which routes exist
4. `navigate()` accepts any `string` — no validation against defined routes
5. `RouteMatch.params` is `Record<string, string>` — untyped
6. `LinkProps.href` is `string` — no validation

## Goal

Thread the route map's literal types through the entire router pipeline:

- `navigate('/nonexistent')` → type error
- `useParams<'/tasks/:id'>()` → `{ id: string }`
- `Link({ href: '/bad' })` → type error
- All changes backward-compatible — unparameterized `Router` defaults to `string`

## Type Flow

```
defineRoutes<const T>() → TypedRoutes<T> → createRouter<T>() → Router<T>
  → navigate(RoutePaths<T>)
  → useRouter<T>() → Router<T>
  → useParams<TPath>() → ExtractParams<TPath>
  → createLink<T>() → Link with href: RoutePaths<T>
```

## Subsumes

- #568 (`useParams()` convenience accessor)
- #569 (simplify Link API)

## Deliverables

1. Design doc with 3 sign-offs (DX, scope, technical)
2. Implementation plan with phased TDD
3. Implementation across 6 phases

PR #576, opened 2026-02-22
docs: type-safe router implementation plan [#572]

viniciusdacal:
## Summary

- 6-phase implementation plan for type-safe router (`@vertz/ui`)
- Design doc: plans/type-safe-router.md (merged via PR #573, 3/3 sign-offs)
- Threads route map literal types through the entire router pipeline

## Phases

1. **POC + Type Utilities** — `PathWithParams<T>`, `RoutePaths<T>`, constraint verification
2. **Generic `defineRoutes<const T>()`** — `TypedRoutes<T>` phantom branded type
3. **`TypedRouter<T>` + typed navigate** — dual overload for contravariance safety
4. **`useRouter<T>()` + `useParams<TPath>()`** — typed context accessors
5. **Typed Link** — `href` validated against `RoutePaths<T>`
6. **Exports, Example Update, Changeset** — public API surface, task-manager example

Each phase includes TDD cycles, integration test acceptance criteria, and a checklist.

## Review request

This is an implementation plan review (not a code PR). Please review for:
- **Technical soundness** — are the TDD cycles in the right order? Any missing edge cases?
- **Phase boundaries** — is each phase independently mergeable?
- **Acceptance criteria** — are they specific enough?

### DIFF
--- a/plans/type-safe-router-impl.md
+++ b/plans/type-safe-router-impl.md
@@ -28,7 +28,7 @@ Phase 1: POC + Type Utilities (PathWithParams, RoutePaths)
 
 ## Phase 1: POC + Type Utilities
 
-**What it implements:** Verifies the `const T extends RouteConfigLike` constraint works (design doc unknown 4.5), then implements `PathWithParams<T>` and `RoutePaths<T>` type utilities.
+**What it implements:** Verifies the `const T extends RouteConfigLike` constraint works (design doc unknown 4.5), then implements `PathWithParams<T>` and `RoutePaths<T>` type utilities. Also creates the E2E acceptance test file (failing — RED state) and the Developer Walkthrough test stub in `packages/integration-tests/`.
 
 **Blocked by:** Nothing — starting phase.
 
@@ -37,6 +37,7 @@ Phase 1: POC + Type Utilities (PathWithParams, RoutePaths)
 1. **POC: `const T extends RouteConfigLike` constraint** — Create a TypeScript Playground or local `.ts` file that verifies:
    - `defineRoutes<const T extends Record<string, RouteConfigLike>>(map: T)` preserves literal keys
    - A concrete loader accessing `params.id` satisfies the constraint
+   - A loader with *no* params access (`() => { return {} }`) satisfies the constraint
    - `RouteConfigLike` with `params: Record<string, string>` doesn't reject valid loaders
    - If POC fails, redesign the constraint before proceeding
 
@@ -46,10 +47,16 @@ Phase 1: POC + Type Utilities (PathWithParams, RoutePaths)
 
 4. **Type tests** in `packages/ui/src/router/__tests__/router.test-d.ts`
 
+5. **Create E2E acceptance test file** at `packages/ui/src/router/__tests__/type-safe-router.test-d.ts` — will fail to compile (RED state) until Phase 5
+
+6. **Create Developer Walkthrough test stub** at `packages/integration-tests/src/__tests__/type-safe-router-walkthrough.test-d.ts` — uses public package imports (`@vertz/ui`), will fail until all phases complete. This catches cross-package type issues per `public-api-validation.md`.
+
 ### Files modified
 
 - `packages/ui/src/router/params.ts` — add `PathWithParams<T>`, `RoutePaths<T>`
 - `packages/ui/src/router/__tests__/router.test-d.ts` — type tests
+- `packages/ui/src/router/__tests__/type-safe-router.test-d.ts` — E2E type acceptance test (failing RED state)
+- `packages/integration-tests/src/__tests__/type-safe-router-walkthrough.test-d.ts` — Developer Walkthrough stub (failing RED state, uses `@vertz/ui` imports)
 
 ### TDD cycles
 
@@ -65,8 +72,14 @@ Phase 1: POC + Type Utilities (PathWithParams, RoutePaths)
 4. **RED:** Type test: `PathWithParams<string>` equals `string` (backward compat)
    **GREEN:** Base case handles `string`
 
-5. **RED:** Type test: `RoutePaths<{ '/': RouteConfig, '/tasks/:id': RouteConfig }>` — `'/tasks/42'` extends it, `'/nonexistent'` does not
+5. **RED:** Type test: `PathWithParams<'/tasks/:id/'>` (trailing slash) equals `` `/tasks/${string}/` ``
+   **GREEN:** Recursive case handles trailing slash naturally
+
+6. **RED:** Type test: `RoutePaths<{ '/': RouteConfig, '/tasks/:id': RouteConfig }>` — `'/tasks/42'` extends it, `'/nonexistent'` does not
    **GREEN:** Implement `RoutePaths<T>` mapped type
+
+7. **RED:** Type test: `RoutePaths<Record<string, never>>` equals `never` (empty route map)
+   **GREEN:** Empty mapped type produces `never` union naturally
 
 ### Integration test acceptance criteria
 
@@ -87,6 +100,10 @@ void _m;
 type _Wild = PathWithParams<'/files/*'>;
 const _w: _Wild = `/files/any/path`; // OK
 void _w;
+
+type _Trail = PathWithParams<'/tasks/:id/'>;
+const _t: _Trail = `/tasks/42/`; // trailing slash preserved
+void _t;
 
 type _Fallback = PathWithParams<string>;
 const _f: _Fallback = 'anything'; // backward compat
@@ -114,14 +131,22 @@ void _invalid;
 type FallbackPaths = RoutePaths<RouteDefinitionMap>;
 const _any: FallbackPaths = '/literally-anything';
 void _any;
+
+// IT-P1-4: RoutePaths of empty map is never
+type EmptyPaths = RoutePaths<Record<string, never>>;
+// @ts-expect-error - never accepts nothing
+const _empty: EmptyPaths = '/anything';
+void _empty;
 ```
 
 ### Acceptance criteria
 
-- [ ] POC confirms `const T extends RouteConfigLike` works with concrete loaders
+- [ ] POC confirms `const T extends RouteConfigLike` works with concrete loaders (with and without params access)
 - [ ] All type tests pass via `bun run typecheck --filter @vertz/ui`
-- [ ] `PathWithParams` handles: static, single param, multi param, wildcard, `string` fallback
-- [ ] `RoutePaths` produces correct union, rejects invalid paths
+- [ ] `PathWithParams` handles: static, single param, multi param, wildcard, trailing slash, `string` fallback
+- [ ] `RoutePaths` produces correct union, rejects invalid paths, empty map → `never`
+- [ ] E2E acceptance test file created (failing — expected at this phase)
+- [ ] Developer Walkthrough stub created in `packages/integration-tests/` with `@vertz/ui` imports (failing — expected)
 - [ ] Lint clean: `bunx biome check packages/ui/src/router/params.ts`
 
 ---
@@ -147,7 +172,7 @@ void _any;
 ### TDD cycles
 
 1. **RED:** Type test: `defineRoutes({ '/tasks/:id': { component: ... } })` — the result's `__routes` should have literal key `'/tasks/:id'`
-   **GREEN:** Add `const T` generic, return `TypedRoutes<T>` with cast
+   **GREEN:** Add `const T` generic, return `TypedRoutes<T>` with cast. Verify all existing runtime tests still pass (regression check — no behavior change).
 
 2. **RED:** Type test: `TypedRoutes<T>` is assignable to `CompiledRoute[]` — must compile when passed to functions expecting `CompiledRoute[]`
    **GREEN:** Intersection type `CompiledRoute[] & { readonly __routes: T }` handles this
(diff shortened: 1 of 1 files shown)


viniciusdacal:
## COMMENT

**Reviewer:** Nora (Technical Feasibility)

Overall this is a well-structured plan. The phase ordering is correct, TDD cycles are mechanical, and the type flow is traced end-to-end. I have no blockers, but several observations that will save implementation time.

---

### TDD Cycle Ordering

**Phase 1 cycles are well-ordered** — negative-first for `PathWithParams` (wildcard, multi-param) drives the recursive template literal branches one at a time. Good.

**Phase 2 cycle 4 is out of place.** The RED step says "Runtime test: existing defineRoutes tests still pass." That is not a RED test — it is a GREEN regression check. Running existing tests to confirm nothing broke belongs in the GREEN step of cycle 3 (after adding the generic), not as a standalone RED/GREEN cycle. Per `tdd.md`, RED means "write ONE failing test that describes a new behavior." Suggestion: fold cycle 4 into cycle 3's GREEN step.

**Phase 3 cycle 5 has the same problem** — "existing createRouter tests still pass" is a regression check, not a RED test.

**Phase 4 cycle ordering is good** — runtime tests for `useParams` come before type tests, which is correct since the runtime implementation must exist before type narrowing matters.

### Phase Boundaries

**All phases are independently mergeable in sequence.** Each modifies different signatures and the type chain builds incrementally. No hidden cross-phase dependencies found.

**One minor concern:** Phase 6 references `examples/task-manager/src/pages/task-detail.tsx` and `examples/task-manager/src/router.ts`, but no `task-manager` example currently exists under `examples/`. The only example is `entity-todo`. Either the task-manager example is expected to be created in a separate PR before Phase 6, or Phase 6 needs to note that it creates these files. Clarify to avoid the implementer stalling.

### Type Flow Gaps

**No dead generics.** The plan traces: `defineRoutes<const T>` -> `TypedRoutes<T>` -> `createRouter<T>` -> `TypedRouter<T>` -> `useRouter<T>()` -> navigate/link. Each step has a positive and negative `.test-d.ts` criterion. This is thorough.

**One gap in the `useRouter<T>` generic plumbing.** The plan says `useRouter<T>()` returns `TypedRouter<T>` via cast. But looking at the current `router-context.ts`, `RouterContext` is `Context<Router>` and `useContext(RouterContext)` returns `Router | undefined`. The plan doesn't specify *how* the cast from `Router` to `TypedRouter<T>` happens inside `useRouter`. The implementation is straightforward (`return useContext(RouterContext) as TypedRouter<T>`), but the plan should note this is the second intentional cast point (the first being `defineRoutes` returning `TypedRoutes<T>`). The TDD rules require documenting all cast points.

**`useRouter<T>()` type parameter ergonomics.** The plan says `useRouter<T>()` takes `T` as the route map type, but the integration tests in Phase 4 (IT-P4-4) show:

```typescript
type T = (typeof typedRoutes)['__routes'];
const _router = useRouter<T>();
```

This is awkward — the developer must extract `__routes` from `TypedRoutes<T>`. The design doc section 1.3 shows `useRouter<typeof routes>()` where `routes` is the return of `defineRoutes()`. But `typeof routes` is `TypedRoutes<T>`, not `T` itself. The generic parameter of `useRouter` needs to accept either `TypedRoutes<T>` directly or provide a utility type like `InferRouteMap<typeof routes>`. The Phase 4 TDD cycles should include a test for the actual developer-facing usage: `useRouter<typeof routes>()` should just work. This is a usability issue that the current plan doesn't address.

### Edge Cases

**Wildcard routes:** Covered in Phase 1 cycle 3 (`PathWithParams<'/files/*'>`) and in the E2E test. Good.

**Multi-param routes:** Covered in Phase 1 cycle 2. Good.

**Empty route maps:** Not tested. `RoutePaths<{}>` should produce `never`. Add a type test: `RoutePaths<Record<string, never>>` equals `never`. This is a minor edge case but prevents surprises.

**Backward compat:** Thoroughly covered — `PathWithParams<string>` = `string`, `RoutePaths<RouteDefinitionMap>` = `string`, `Router` (no generic) accepts any string, `useRouter()` with no param returns `Router`. All tested. Good.

**Trailing slash paths:** Not addressed. `PathWithParams<'/tasks/:id/'>` (trailing slash) will produce `` `/tasks/${string}/` `` which may not match `'/tasks/42'`. This is arguably correct behavior, but worth a type test to lock it down either way.

### Acceptance Criteria

**Specific and testable throughout.** Each phase has concrete `@ts-expect-error` negative tests and positive compilation checks. The integration test code blocks are copy-pasteable.

**Phase 6 acceptance criteria could be tighter.** "Example compiles with `useParams<'/tasks/:id'>()`" is vague if the example doesn't exist yet (see above). Should be: "Example file created/updated, `bun run typecheck` passes with the example in scope."

### Risk Areas

**`const T` constraint (Unknown 4.5):** The plan correctly gates this behind a POC in Phase 1, subtask 1. The proposed `RouteConfigLike` with `params: Record<string, string>` is sound — a concrete `{ id: string }` extends `Record<string, string>`, so the constraint won't reject valid loaders. The current `RouteConfig` uses `ExtractParams<TPath>` which evaluates to `Record<string, never>` for `TPath = string` — that's the root of the constraint conflict. The fix is correct. Just make sure the POC also tests a loader with *no* params access (the `() => { return {} }` case) — the constraint must not *require* params usage.

**Contravariance dual overload (Design Doc 4.4):** The dual overload approach is sound. `TypedRouter<T>` with `{ (url: RoutePaths<T>): ...; (url: string): ... }` is assignable to `Router` via the second overload, and TypeScript resolves to the first (narrow) overload at call sites. One subtlety: if `RoutePaths<T>` evaluates to `string` (which it does when `T = RouteDefinitionMap`), both overloads accept `string`, and TypeScript picks the first. This means `TypedRouter<RouteDefinitionMap>` behaves identically to `Router`. Good — no behavioral difference in the default case.

**Phantom branded types (`TypedRoutes<T>`):** The intersection `CompiledRoute[] & { readonly __routes: T }` is the standard phantom pattern. The risk is that `.map()`, `.filter()`, spread, etc. strip the brand (producing `CompiledRoute[]`). The design doc section 9.3 acknowledges this. The plan should add a type test: `[...typedRoutes]` produces `CompiledRoute[]` (not `TypedRoutes<T>`), to lock this down as intentional behavior. This prevents future confusion.

### Suggestions (nice-to-have)

1. **Add a `PathWithParams<'/tasks/:id/'>` trailing slash test** in Phase 1 to document the behavior.

2. **Add an empty route map test** (`RoutePaths<Record<string, never>>`) in Phase 1.

3. **Add a brand-stripping test** (`typeof [...typedRoutes]` is `CompiledRoute[]`) in Phase 2.

4. **Fold "existing tests still pass" cycles** (Phase 2 cycle 4, Phase 3 cycle 5) into the GREEN step of their preceding cycle. They are not RED tests.

5. **Clarify `useRouter<T>()` ergonomics** — should the generic accept `TypedRoutes<T>` (the `typeof routes` that developers actually have) or the raw map `T`? The current plan is inconsistent between the design doc examples and the implementation test code.

6. **Clarify the task-manager example** in Phase 6 — does it already exist, or does Phase 6 create it?

7. **Consider adding `type-safe-router.test-d.ts` as a failing file in Phase 1** (the E2E acceptance test from the design doc). The plan mentions "Created in Phase 1 as a failing test, passes after Phase 5" in the E2E section, but Phase 1's subtask list doesn't include creating this file. Make it explicit in Phase 1 subtasks.

viniciusdacal:
## APPROVE

**Reviewer:** Nora (Technical Feasibility) -- follow-up review

All 7 of my original feedback points have been addressed. Here is the point-by-point verification, plus observations on the updated plan.

---

### Original feedback — status

**1. Phase 2 cycle 4 and Phase 3 cycle 5 regression checks should be folded into GREEN steps, not standalone TDD cycles**

Addressed. Phase 2 now has 4 cycles, and the old "existing tests still pass" cycle has been replaced with a proper RED test for brand-stripping on array spread (cycle 4). The regression check ("Verify all existing runtime tests still pass") is now folded into the GREEN step of cycle 1, exactly where it belongs. Phase 3 similarly folds the regression check into cycle 1's GREEN step and replaces the old cycle 5 with a proper backward-compat test (cycle 4: unparameterized `Router` accepts any string navigate). Both phases now have only proper RED/GREEN cycles. Good.

**2. `useRouter<T>()` cast point not documented; generic param ergonomics inconsistent**

Addressed. Phase 4 now has a dedicated "Cast point" section that explicitly documents `useRouter<T>()` as the second intentional cast point and explains why the cast is safe. The ergonomics issue is resolved with `InferRouteMap<T>` utility type — developers write `useRouter<InferRouteMap<typeof routes>>()` instead of extracting `__routes` manually. The recommended `useAppRouter()` wrapper is documented. The integration test (IT-P4-4) now demonstrates the actual developer-facing usage with `InferRouteMap`. This is a solid improvement. One minor note: `InferRouteMap<typeof routes>` is still a bit verbose compared to the design doc's `useRouter<typeof routes>()` — but the plan mentions the implementation should "choose whichever approach requires the least ceremony at the call site," which leaves room for the implementer to optimize. Acceptable.

**3. Missing edge cases: empty route maps, trailing slashes, brand-stripping on array spread**

All three addressed.
- **Empty route maps:** Phase 1 cycle 7 tests `RoutePaths<Record<string, never>>` equals `never`. Also appears in integration test IT-P1-4.
- **Trailing slashes:** Phase 1 cycle 5 tests `PathWithParams<'/tasks/:id/'>` equals `` `/tasks/${string}/` ``. Also appears in integration test IT-P1-1.
- **Brand-stripping on array spread:** Phase 2 cycle 4 is now a dedicated RED/GREEN test for `[...typedRoutes]` producing `CompiledRoute[]`. Integration test IT-P2-3 covers this. The acceptance criteria explicitly say "tested + documented as intentional."

**4. Phase 6 references `examples/task-manager/` — needs clarification if it exists**

Addressed. The `examples/task-manager/` directory exists (I verified: it contains `router.ts`, `pages/task-detail.tsx`, and the expected structure). Phase 6 now has detailed subtasks (subtasks 3 and 4) that specify exactly what changes to make in `router.ts` (remove `: Router` annotation, add `useAppRouter()` export, wire `createLink<T>()`) and `task-detail.tsx` (replace `router.current.value?.params.id` with `useParams`, replace `useRouter()` with `useAppRouter()`). The code examples in the subtask descriptions are concrete and copy-pasteable. This is much better than the vague "verify type inference flows" from v1.

**5. E2E test file not explicitly in Phase 1 subtasks**

Addressed. Phase 1 subtask 5 now explicitly says: "Create E2E acceptance test file at `packages/ui/src/router/__tests__/type-safe-router.test-d.ts` — will fail to compile (RED state) until Phase 5." It also appears in the "Files modified" list for Phase 1. Clear and unambiguous.

**6. POC should also test loader with no params access**

Addressed. Phase 1 subtask 1 now includes: "A loader with *no* params access (`() => { return {} }`) satisfies the constraint" as a bullet point in the POC verification checklist. This ensures the `RouteConfigLike` constraint does not *require* params usage.

**7. `type-safe-router.test-d.ts` creation should be explicit in Phase 1**

Addressed. Same as point 5 — subtask 5 and the files modified list both call this out. Additionally, subtask 6 adds a Developer Walkthrough test stub in `packages/integration-tests/` with public package imports, which aligns with `public-api-validation.md`. Both files are listed in the "Files modified" section. Good.

---

### Other reviewer feedback also addressed

I also verified that the other reviewer's (josh's) feedback was incorporated:

- **`useAppRouter()` in example** (josh #1): Phase 6 subtask 3 adds `useAppRouter()` export to `router.ts`, subtask 4 shows `TaskDetailPage` using it.
- **`useParams` error message** (josh #3): Phase 4 TDD cycle 1 specifies a `useParams`-specific error message (`"useParams() must be called within RouterContext.Provider"`), and subtask 3 says it uses "its own null check and `useParams`-specific error message." Phase 4 acceptance criteria explicitly list this.
- **Remove `: Router` annotation** (josh #4): Phase 6 subtask 3 explicitly says "Remove the `: Router` annotation on `appRouter`" with an explanation of why it erases the typed inference.
- **Wire `createLink<T>()`** (josh #5): Phase 6 subtask 3 includes a concrete code example for `createLink<InferRouteMap<typeof routes>>()`.
- **`ui-components.md` timing** (josh #7): Now updated in Phase 4 (subtask 6) for `useParams`, and again in Phase 6 (subtask 5) for `useAppRouter`. The acceptance criteria for Phase 4 explicitly include the doc update.
- **Subpath exports test completeness** (josh #13, #14): Phase 6 subtask 2 explicitly lists all three updates: `expectedExports` array, "same references" test, and "main barrel re-exports" test. The integration test criteria IT-P6-1 through IT-P6-3 validate each.
- **Changeset backward-compat note** (josh #15): Phase 6 subtask 6 includes a concrete changeset description mentioning "Fully backward-compatible."
- **Post-implementation retrospective** (josh #16): Phase 6 subtask 9 adds this with the expected file path.
- **Type-only exports** (josh #12): Phase 6 subtask 1 explicitly separates runtime (`useParams`) from type-only (`export type { TypedRouter, TypedRoutes, RoutePaths, PathWithParams, InferRouteMap }`).

---

### New observations on the updated plan

Nothing blocking. Two minor observations:

**1. Phase 4 `InferRouteMap` file location is ambiguous.** Subtask 1 says "in `packages/ui/src/router/define-routes.ts` (or `params.ts`)" — the implementer should pick one and be consistent. Since `InferRouteMap` depends on `TypedRoutes<T>` (which lives in `define-routes.ts`), co-locating it in `define-routes.ts` is the natural choice. Not a blocker — the implementer will sort this out.

**2. The Developer Walkthrough test (Phase 1 subtask 6) content is described by reference** — "Same test content as the internal test above, but using public package imports." This is clear enough for the implementer, but note that the walkthrough test may need slightly different imports (e.g., `InferRouteMap` from `@vertz/ui` instead of `../define-routes`). The Phase 6 E2E section already shows the correct public imports. No action needed — just a heads-up for the implementer.

---

### Verdict

**APPROVE.** All 7 original points addressed adequately. The plan is thorough, TDD-compliant, and ready for implementation. The phase ordering is correct, the type flow is fully traced, and the edge cases are covered. No dead generics, no missing acceptance criteria, no process gaps.

Issue #510, opened 2026-02-21
feat(ui): tolerant hydration mode for mount() — scoped diff with browser extension tolerance


viniciusdacal:
## Summary
Implement `'tolerant'` and `'strict'` hydration modes for mount(). Currently mount() defaults to `'replace'` which clears SSR content and re-renders, causing a flash of empty content.

**This is P0 immediately post-announcement.** The `'replace'` mode is unacceptable for production apps with SSR.

Design doc: `plans/mount-api-design.md`

### Requirements
- Scoped to root container only
- Extra DOM nodes (browser extensions) are ignored, not flagged as errors
- Missing nodes are re-created
- Attribute mismatches are patched silently
- Dev mode: console.warn for mismatches
- Prod mode: silent

### Technical Challenge
Current hydration is marker-based (per-component). Tolerant hydration requires tree diffing against rendered output. This may need a lightweight reconciler or a different approach.

### Research Needed
- Can we extend the marker-based system to handle tolerance?
- Do we need a minimal vdom/reconciler?
- What do Astro, Qwik, and Solid do for hydration tolerance?

### Definition of Done
- `'tolerant'` mode works as default
- `'strict'` mode available for apps that need guaranteed consistency
- No browser extension mismatch errors
- No flash of empty content
- Performance benchmarks vs 'replace' mode

PR #10, opened 2026-02-05
docs: @vertz/testing implementation plan

viniciusdacal:
## Summary
- Implementation plan for `@vertz/testing` — the testing utilities package for the Vertz framework
- Defines the complete architecture: TestAppBuilder (builder pattern), RequestBuilder (thenable), MockRegistry (two-scope mocking), ResponseParser/Validator, TestServiceBuilder (unit testing services), and the `vertz.testing` namespace re-export
- Follows the same format as the core implementation plan: Architectural Decisions, Package Structure, Key Implementation Details with code snippets, package.json, 12 TDD phases, Verification checklist, Open Items

## Test plan
- [ ] Review plan for completeness against testing design (`vertz-testing-design.md`)
- [ ] Verify TDD phases are ordered correctly (foundation types → request factory → response → mocks → builders → integration)
- [ ] Check alignment with core implementation plan conventions (bunup, vitest globals, ESM, isolatedDeclarations)
- [ ] Verify mock scoping design (app-level + per-request overrides, deep merge for services, replace for middleware)
- [ ] Confirm the `vertz` namespace re-export pattern matches core implementation plan

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #619, opened 2026-02-22
feat(ui): add tolerant hydration mode for mount() [#510]

viniciusdacal:
## Summary

Implements tolerant hydration mode for `mount()` ([#510](https://github.com/vertz-dev/vertz/issues/510)). When SSR HTML exists in the DOM, `mount(App, '#root', { hydration: 'tolerant' })` walks the existing nodes and attaches reactivity instead of clearing and re-rendering — eliminating the flash of empty content.

### Key changes

- **Hydration context** (`hydration-context.ts`) — Cursor-based DOM walker with claim functions that advance through SSR nodes, skipping foreign elements (browser extensions, ad blockers)
- **DOM helper hydration branches** — `__element`, `__text`, `__child`, `__insert` adopt existing SSR nodes when hydrating; new `__append` (no-op during hydration), `__staticText`, `__enterChildren`/`__exitChildren`
- **Conditional + list hydration** — `__conditional` claims comment anchors and active branch content; `__list` claims existing items and skips initial reconciliation
- **Mount integration** — Tolerant mode with error recovery (failed hydration → automatic fallback to CSR replace mode), dev warnings for empty root and mismatches
- **Compiler changes** — Emits `__enterChildren`/`__exitChildren` around child construction, replaces `appendChild` → `__append`, `createTextNode` → `__staticText`

### Error recovery

| Scenario | Behavior |
|----------|----------|
| Browser extension node encountered | Skipped silently (`console.debug` in dev) |
| Expected node not found | Created new element + `console.warn` in dev |
| Tolerant mode on empty root | Warning + fallback to replace mode |
| Hydration throws exception | `endHydration()` + fallback to full CSR re-render |

## Test plan

- [x] Hydration context unit tests (14 tests) — claim functions, cursor stack, enter/exit nesting
- [x] DOM helper hydration tests (19 tests) — adopt elements, text nodes, reactive effects, __append no-op
- [x] Conditional hydration tests (4 tests) — comment anchor, branch content, reactive switching
- [x] List hydration tests (3 tests) — item claiming, skip reconciliation, post-hydration updates
- [x] Mount hydration tests (9 tests) — SSR preservation, extension tolerance, event handlers, error recovery
- [x] E2E integration test — SSR HTML → tolerant mount → verify same DOM references → click handler + reactive text → extension node preserved
- [x] All 984 existing UI tests pass (zero regressions)
- [x] All 276 compiler tests pass (zero regressions)
- [x] TypeScript: both packages clean
- [x] Biome lint: clean

### DIFF
--- a/packages/ui/src/dom/__tests__/hydration-element.test.ts
+++ b/packages/ui/src/dom/__tests__/hydration-element.test.ts
@@ -95,6 +95,19 @@ describe('DOM helpers — hydration branches', () => {
       expect(wrapper).toBe(span);
     });
 
+    it('creates new wrapper when claim fails during hydration', () => {
+      const root = document.createElement('div');
+      // Only a text node, no span to claim
+      root.appendChild(document.createTextNode('just text'));
+      startHydration(root);
+
+      const wrapper = __child(() => 'fallback');
+      // Should have created a new span wrapper since no span was found
+      expect(wrapper.tagName).toBe('SPAN');
+      expect(wrapper.style.display).toBe('contents');
+      expect(wrapper.textContent).toBe('fallback');
+    });
+
     it('attaches reactive effect to adopted wrapper', () => {
       const root = document.createElement('div');
       const span = document.createElement('span');
@@ -170,6 +183,18 @@ describe('DOM helpers — hydration branches', () => {
     it('creates new text node during CSR', () => {
       const text = __staticText('hello');
       expect(text.data).toBe('hello');
+    });
+
+    it('creates new text node when claim fails during hydration', () => {
+      const root = document.createElement('div');
+      // Only an element node, no text nodes to claim
+      root.innerHTML = '<span></span>';
+      startHydration(root);
+
+      const text = __staticText('fallback');
+      // Should have created a new text node since no text node was found
+      expect(text.data).toBe('fallback');
+      expect(text).not.toBe(root.firstChild);
     });
   });
 
--- a/packages/ui/src/hydrate/__tests__/hydration-context.test.ts
+++ b/packages/ui/src/hydrate/__tests__/hydration-context.test.ts
@@ -238,6 +238,21 @@ describe('hydration-context', () => {
     });
   });
 
+  describe('exitChildren bounds checking', () => {
+    it('warns in dev mode when called with empty stack', () => {
+      const root = document.createElement('div');
+      root.innerHTML = '<div></div>';
+      startHydration(root);
+
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+      exitChildren(); // No matching enterChildren — stack is empty
+      expect(warnSpy).toHaveBeenCalledWith(
+        expect.stringContaining('exitChildren() called with empty stack'),
+      );
+      warnSpy.mockRestore();
+    });
+  });
+
   describe('concurrent hydration guard', () => {
     it('throws if startHydration is called while already hydrating', () => {
       const root1 = document.createElement('div');
--- a/packages/ui/src/hydrate/hydration-context.ts
+++ b/packages/ui/src/hydrate/hydration-context.ts
@@ -140,5 +140,15 @@ export function enterChildren(el: Element): void {
  * Called by compiler-emitted `__exitChildren()`.
  */
 export function exitChildren(): void {
+  if (cursorStack.length === 0) {
+    if (typeof process !== 'undefined' && process.env.NODE_ENV !== 'production') {
+      console.warn(
+        '[hydrate] exitChildren() called with empty stack. ' +
+          'This likely means __exitChildren was called without a matching __enterChildren.',
+      );
+    }
+    currentNode = null;
+    return;
+  }
   currentNode = cursorStack.pop() ?? null;
 }


viniciusdacal:
## Developer Experience Review — Josh

### API Concerns

**1. `'strict'` in the type union is a footgun waiting to fire**

The type `hydration?: 'replace' | 'tolerant' | 'strict'` accepts `'strict'` today, but passing it silently falls through to replace mode. There is no runtime error, no dev warning, nothing. A developer who reads the type definition and writes `{ hydration: 'strict' }` thinking they're getting strict hydration will get replace mode instead — the exact opposite of what they intended.

This violates "Explicit over implicit" directly. Either remove `'strict'` from the union until it's implemented, or throw a clear error:

```ts
if (mode === 'strict') {
  throw new Error(
    "mount(): hydration: 'strict' is reserved but not yet implemented. " +
    "Use 'tolerant' for SSR hydration or 'replace' (default) for CSR."
  );
}
```

An AI agent reading the TypeScript types would absolutely try `'strict'` and get silently incorrect behavior. This directly conflicts with Vision principle 3 ("Can an LLM use this correctly on the first prompt?").

**2. `'tolerant'` is an unusual word choice — consider the discoverability angle**

When I think "hydration mode," I think of words like `'reuse'`, `'attach'`, `'adopt'`, or `'resume'`. The word `'tolerant'` communicates *fault tolerance* (which is accurate — it tolerates browser extension nodes), but a developer coming from React/Solid/Svelte might not guess that `'tolerant'` means "walk the existing DOM and attach reactivity."

Counterpoint: the design doc explains the rationale clearly, and once you learn the word, it's distinctive and memorable. The inline JSDoc `'tolerant' (walk SSR DOM)` helps. I'd say this is acceptable if the docs are good, but worth noting that `'hydrate'` or `'adopt'` would be more self-describing for the core behavior. Not a blocker — just a discoverability consideration.

**3. The old `hydration?: 'replace' | false` was replaced — verify no external consumers**

The PR changes the `MountOptions.hydration` type from `'replace' | false` to `'replace' | 'tolerant' | 'strict'`. This removes `false` from the union. If any existing code passes `hydration: false`, it now gets a type error. The changeset marks this as a `patch`, which is fine under the pre-v1 semver policy, but the changeset description should mention the breaking type change for anyone upgrading.

**4. `mount()` vs `hydrate()` naming could confuse newcomers**

The design doc is explicit that these are mutually exclusive APIs for different architectures. But from the public API surface alone (`index.ts`), a developer sees both `mount` and `hydrate` exported side-by-side and the word "hydration" in `MountOptions`. There's no JSDoc on `mount()` that says "for full-app hydration, not to be confused with `hydrate()` for islands." An AI agent scanning exports would likely conflate them.

Suggestion: Add a brief JSDoc note to `mount()` distinguishing it from `hydrate()`, and vice versa. Something like:

```ts
/**
 * Mount an app to a DOM element.
 *
 * For full-app SSR hydration, use `{ hydration: 'tolerant' }`.
 * For island/per-component hydration, use `hydrate()` instead.
 */
```

### Error Message Issues

**5. "Falling back to replace mode" — is this scary or reassuring?**

The warning `[mount] Hydration failed, falling back to replace mode:` is technically correct, but "falling back" in production contexts often means "something went wrong and we're doing something worse." A developer seeing this in their console might panic.

Suggestion: reword to make the recovery explicit:

```
[mount] Hydration failed — re-rendering from scratch (no data loss). Error:
```

This communicates the same information but frames the outcome positively.

**6. "Did you mean 'replace'?" in the empty-root warning is presumptuous**

The message `hydration: "tolerant" used on empty root. Did you mean "replace"?` assumes the developer made a mistake. But what if they're developing locally without SSR and want to test the tolerant code path? The "Did you mean" phrasing feels condescending.

Better:

```
[mount] hydration: "tolerant" has no effect on an empty root (no SSR content found). Using replace mode.
```

This states the fact and the consequence without guessing intent.

**7. `claimElement` skips ALL non-matching elements with only `console.debug`**

If the SSR output has a real mismatch (developer renders `<div>` on server but `<section>` on client), `claimElement` will skip the server's `<div>` as if it were a browser extension node, logging only a `debug` message. The developer gets no useful feedback about their actual bug — the mismatch is silently swallowed.

The current logic cannot distinguish "browser extension node" from "developer's SSR/client mismatch." Both are non-matching elements. This is a known tradeoff of the tolerant approach, but the debug message `Skipping foreign node: <div>` would be actively misleading in the mismatch case — `<div>` is clearly not a "foreign node."

Suggestion: Change the debug message to be more neutral:

```
[hydrate] Skipping non-matching node: <div> (expected <section>)
```

This helps developers spot real mismatches without changing the tolerant behavior.

**8. No warning when hydration "succeeds" but leaves unclaimed nodes**

If the SSR output has extra nodes that the app doesn't claim (not browser extensions, but actual content the client doesn't render), hydration completes silently. The developer has no way to know their server and client trees diverged. Consider a dev-mode `console.debug` at `endHydration()` if the cursor hasn't reached the end of siblings.

### Manifesto Alignment

**9. "Explicit over implicit" — strong alignment, one gap**

The opt-in `{ hydration: 'tolerant' }` is excellent. No auto-detection magic. The developer makes a conscious choice. This nails the principle.

The gap: the silent `'strict'` fallthrough (point 1 above). Accepting a value and ignoring it is the definition of implicit behavior.

**10. "One way to do things" — solid**

Full-app hydration: `mount()` with `'tolerant'`. Island hydration: `hydrate()`. No overlap, no ambiguity. The design doc is clear about mutual exclusivity. This is well done.

**11. "AI agents are first-class users" — mostly good, two issues**

An AI agent reading the TypeScript types and JSDoc could correctly use `mount(App, '#root', { hydration: 'tolerant' })`. The API is self-describing.

Issues for AI agents:
- The `'strict'` value in the union would mislead an agent into thinking it's functional (point 1)
- The `mount()` vs `hydrate()` naming overlap requires reading docs beyond the type signatures (point 4)

**12. "Compile-time over runtime" — acceptable tradeoff**

The `getIsHydrating()` check runs at runtime on every `__element`, `__text`, `__append`, etc. call. This is a function call + boolean check per DOM operation, which is minimal overhead. The real work (cursor walking) only happens when hydration is active. The compiler changes (`__enterChildren`/`__exitChildren`) push as much structure as possible to build time.

The one thing that could be a compile-time check: if the compiler knows the app is SSR-only (no hydration), it could skip emitting `__enterChildren`/`__exitChildren` entirely. This is a future optimization, not a blocker.

### Migration & Workflow

**13. Migration path is clean**

Going from `mount(App, '#root')` to SSR+hydration requires exactly one change: add `{ hydration: 'tolerant' }`. The app function, component code, and everything else stays the same. The compiler changes (`__append` instead of `appendChild`) are transparent to developers. This is excellent.

**14. Accidental `'tolerant'` without SSR is handled well**

Empty root + `'tolerant'` = dev warning + automatic fallback to replace mode. The app still works. This is the right call — fail gracefully, warn loudly.

**15. Error recovery is solid**

If hydration throws, the catch block calls `endHydration()`, warns, and falls back to replace mode. The app recovers. The developer sees the error. This is good defensive design.

One edge case: if `app()` throws during the hydration attempt and the root DOM is left in a partially-claimed state, the fallback `root.textContent = ''` clears everything and starts fresh. This is correct but worth noting in docs — hydration failure means a full re-render flash, which is the same as not using hydration at all.

### Minor DX Issues

**16. `enterChildren` is exported from `hydration-context.ts` as a public name but also wrapped as `__enterChildren` in `element.ts`**

The hydration-list test imports `enterChildren` directly from `hydration-context.ts`:

```ts
import { endHydration, enterChildren, startHydration } from '../../hydrate/hydration-context';
```

This means `enterChildren` (without the `__` prefix) is used as an internal API in tests, while `__enterChildren` (with prefix) is the compiler-facing export. This is fine architecturally, but the naming inconsistency (no `__` prefix for the context function, `__` prefix for the wrapper) could confuse someone reading the test code. A comment in the test clarifying "using the internal cursor API directly, not the compiler wrapper" would help.

**17. The `globals.d.ts` for `process` is a pragmatic solution but worth documenting**

The new `globals.d.ts` declares `process.env.NODE_ENV` for dev-mode guards. This is standard practice but not documented anywhere in the PR. A one-line comment in the file explaining "bundlers replace this at build time; the `typeof process` guard prevents runtime errors in browsers" would help future maintainers. (The file does have this comment — good.)

**18. The changeset description is dense but complete**

The changeset in `.changeset/tolerant-hydration-mount.md` packs a lot into one paragraph. It's accurate but hard to scan. Breaking it into bullet points would improve readability for changelog consumers.

### What's Good

- **The cursor-based walk-and-attach approach is elegant.** No virtual DOM, no reconciler, no diffing. Just walk the tree in order and attach reactivity. This is the minimum runtime for the problem.

- **Error recovery with automatic CSR fallback is the right default.** Hydration is hard, and the graceful degradation to replace mode means a broken SSR setup doesn't break the app — it just shows a flash.

- **The `__append` no-op pattern is clever.** By replacing `appendChild` with a function that's a no-op during hydration, the same compiled output works for both CSR and SSR hydration. No conditional compilation needed.

- **Browser extension tolerance is well-tested.** The e2e test injects a `grammarly-extension` node and verifies it survives hydration. This is exactly the real-world scenario that motivates the feature.

- **Test coverage is thorough.** 14 context tests, 19 DOM helper tests, 4 conditional tests, 3 list tests, 9 mount tests, plus the e2e integration. The tests read well and cover both hydration and CSR paths.

- **The design doc is one of the best I've seen for this codebase.** Clear context, manifesto alignment, explicit non-goals, phased implementation, concrete test lists. If every feature had a design doc this thorough, the codebase would be in great shape.

- **The `__staticText` / `__text` separation is clean.** Static text (no reactivity) gets a simpler code path. Reactive text gets the effect. The compiler makes the right choice, and the developer never has to think about it.

### Verdict

**APPROVE with minor requested changes:**

1. **Must fix:** Either remove `'strict'` from the type union or throw an error when it's passed. Silent acceptance of an unimplemented mode is a DX bug. (Point 1)
2. **Should fix:** Improve the `claimElement` debug message to include the expected tag — helps developers distinguish real mismatches from extension tolerance. (Point 7)
3. **Should fix:** Add JSDoc to `mount()` and `hydrate()` distinguishing the two APIs. (Point 4)
4. **Nice to have:** Reword "Falling back to replace mode" and "Did you mean 'replace'?" messages. (Points 5, 6)

The core design is sound, the implementation is clean, and the test coverage is excellent. The issues above are about polishing the developer-facing surface — the kind of things that prevent frustrated GitHub issues six months from now.

viniciusdacal:
## Adversarial Review: Tolerant Hydration Mode (Josh)

Found several issues worth addressing:

### Critical Issues

**1. `'strict'` mode is implemented but not functional**
- Type allows `hydration?: 'replace' | 'tolerant' | 'strict'`
- Only `'tolerant'` is handled; `'strict'` silently falls back to `'replace'`
- No validation or warning for unsupported mode

```typescript
// mount.ts line 73 - no handling for 'strict'
if (mode === 'tolerant') { ... }
// Anything else (including 'strict') falls through to replace
```

**2. Potential memory leak in `__child` effect during hydration**
- In `element.ts`, the effect attaches but the cleanup path isn't fully clear
- If the wrapper is disposed while hydrating, the effect scope might not clean up properly

### Edge Cases & Bugs

**3. `exitChildren` has no bounds checking**
```typescript
// hydration-context.ts
export function exitChildren(): void {
  currentNode = cursorStack.pop() ?? null;
}
```
If `__exitChildren` is called without matching `__enterChildren`, it silently returns null. Should throw or warn in dev.

**4. `__conditional` hydration: branch content may not be properly claimed**
- The test shows it creates a fragment, but real SSR output may have siblings
- No validation that the claimed nodes actually belong to the conditional branch
- If SSR HTML has extra whitespace or comments, hydration cursor could drift

**5. Browser compatibility: `process.env.NODE_ENV` guards**
- `globals.d.ts` declares `process`, but runtime behavior varies by bundler
- Some aggressive tree-shaking configs may strip these checks

### Performance Concerns

**6. No hydration batch/defer mechanism**
- Each `__text`, `__element`, `__child` creates individual effects
- Large DOM trees will trigger many micro-tasks

**7. List hydration runs all render functions synchronously**
For 1000+ items, this blocks the main thread.

### Accessibility

**8. No ARIA attribute validation during adoption**
When `__element` adopts an existing SSR node, mismatched ARIA attributes could cause a11y issues.

### Testing Gaps

**9. Missing test: `__child` hydration with reactive content**
- Current test uses static `__child(() => 'hello')`
- Need test for reactive signal values

**10. Missing test: mismatched element type during hydration**
- What happens if SSR has `<div>` but component expects `<span>`?

### Minor Issues

**11. Dev warnings go to `console.debug` for skipped extension nodes**
- `console.debug` is often hidden in production browsers
- Consider `console.warn` or `console.info`

**12. No way to disable hydration logging**

---

**Recommendation:** Address items 1, 3, and 4 before merge. Items 5-7 are nice-to-have improvements. Items 8-12 can be follow-ups.

🤖 Reviewed by josh (Developer Relations)

viniciusdacal:
## Re-review requested — all findings addressed

All 21 review findings from the initial review round have been addressed across two follow-up commits:

### Commit `57d3202` — First round of fixes
- **Critical #1**: Fixed `hydrateConditional` ripping SSR nodes from DOM (DocumentFragment bug)
- **Critical #2**: `'strict'` mode now throws explicit error instead of silent fallthrough
- **#5**: Added concurrent hydration guard in `startHydration()`
- **#7**: Rewrote conditional hydration tests to use `__element`/`__staticText`/`__append`
- **#8**: Added list e2e test through `mount()` (SSR items adopted + reactive updates)
- **#10**: Improved `claimElement` debug message to include expected tag
- **#11**: Added JSDoc to `mount()` distinguishing full-app vs island hydration
- **#12**: Reworded warning messages for clarity and actionability
- **#20**: Added test for `exitChildren` on empty elements
- **#4**: Added 4 compiler tests (nested enter/exit, childless, fragments, imports)

### Commit `42a75ae` — Second round of fixes
- **#6**: Stale effects cleanup — `app()` wrapped in disposal scope; `runCleanups()` on error recovery
- **#13**: Added `mount.test-d.ts` type tests for `MountOptions.hydration` (positive + `@ts-expect-error` negative)
- **#15**: Added list nodeMap reorder test (adopted SSR nodes reused after hydration)
- **#14**: Updated design doc with implementation deviations (`process.env.NODE_ENV`, disposal scope, strict guard)
- **#16**: Retrospective written at `plans/post-implementation-reviews/tolerant-hydration.md`

### Quality gates
- **992 UI tests** ✅
- **280 compiler tests** ✅
- **Typecheck** ✅
- **Lint** ✅
- **All 60 turbo quality gates** ✅

Ready for re-review.

Issue #632, opened 2026-02-23
Mutation testing: improve test health across packages


viniciusdacal:
# Mutation Testing: Test Health Improvement Tracking

## Background

We ran a mutation testing spike using [Stryker Mutator](https://stryker-mutator.io/) on `@vertz/errors` and `@vertz/core` to assess test quality. Mutation testing works by making small changes (mutants) to source code and checking if tests catch them. Surviving mutants = weak tests.

## Results Summary

| Package | Mutants | Score (total) | Score (covered) |
|---------|---------|---------------|-----------------|
| `@vertz/errors` | 462 | **74.68%** | 87.12% |
| `@vertz/core` | 762 | **70.87%** | 82.32% |

## Strongest Files (100% mutation score)

| Package | File |
|---------|------|
| errors | `domain/auth.ts`, `domain/client.ts`, `result.ts`, `app-error.ts` |
| core | `env/env-validator.ts`, `response-utils.ts`, `http-exceptions.ts`, `freeze.ts`, `make-immutable.ts`, `module-def.ts`, `router-def.ts`, `service.ts`, `vertz.ts` |

## Weakest Files

| Package | File | Score | Problem |
|---------|------|-------|---------|
| errors | `domain/migration.ts` | **0%** | Zero test coverage (26 mutants) |
| errors | `domain/schema.ts` | **0%** | Zero test coverage (8 mutants) |
| core | `app/bun-adapter.ts` | **0%** | Zero test coverage (8 mutants) |
| core | `app/app-builder.ts` | **19%** | `listen()`, entity routes untested |
| errors | `fetch.ts` | **34%** | HTTP error subclasses and `createHttpError()` untested |
| core | `server/cors.ts` | **64%** | Default headers, credentials, exposed headers not verified |
| core | `app/app-runner.ts` | **54%** | Error handling paths survive mutation |

## Three Systemic Weaknesses

1. **Type guards only tested positively** — `isUniqueViolation()` etc. only tested with matching errors. Mutating to `return true` passes all tests.
2. **Error code/name strings never asserted** — constructor strings like `'BadRequest'`, `'NetworkError'` never checked.
3. **Boolean conditionals lack boundary tests** — `body === null || body === undefined || body === ''` can lose clauses without test failure.

### Wave 1 — Zero-coverage files (parallel)
- [ ] `errors/domain/migration.ts` — #633
- [ ] `errors/domain/schema.ts` — #634
[...]

PR #638, opened 2026-02-23
test(core): add tests for app/bun-adapter.ts

### LINKED ISSUE #635: test(core): add tests for app/bun-adapter.ts (0% mutation score)
## Description

`packages/core/src/app/bun-adapter.ts` has **0% mutation score** — zero test coverage. All 8 mutants survive. This was identified in the mutation testing spike (#632).

## What to test

Read `packages/core/src/app/bun-adapter.ts` and write tests covering:

1. **Bun server adapter creation** — verify it produces a valid server configuration
2. **Request handling** — verify incoming Bun requests are correctly converted to the framework's internal format
3. **Response handling** — verify framework responses are correctly converted to Bun responses
4. **Port/host configuration** — verify adapter respects configuration options
5. **Edge cases** — empty body, missing headers, etc.

## Acceptance Criteria

- [ ] Every exported function/method has at least one test
- [ ] Both success and error paths are tested
- [ ] Tests pass: `bun run --filter @vertz/core test`
- [ ] Mutation score for `app/bun-adapter.ts` reaches >80%

## Files

- **Source:** `packages/core/src/app/bun-adapter.ts`
- **Tests:** `packages/core/src/app/__tests__/bun-adapter.test.ts` (create new)
- **Reference:** Check existing test patterns in `packages/core/src/` for conventions

## Constraints

- Do NOT modify the source file — only add tests
- Follow existing test patterns in `packages/core/`
- This may require mocking Bun server APIs — keep mocks minimal and focused
- TDD: write failing tests first, verify they fail for the right reason, then confirm they pass

Part of #632

viniciusdacal:
## Summary
- Add 8 tests for `packages/core/src/app/bun-adapter.ts` which had 0% mutation score (8 surviving mutants)
- Tests verify the adapter's observable contracts: port/hostname forwarding, server handle properties, close behavior
- Tests cover edge cases: omitted hostname, random port binding (port 0)
- Uses minimal Bun.serve mock — tests actual adapter logic, not Bun internals

## Test plan
- [x] All tests pass: `bun run --filter @vertz/core test`
- [x] Only adds a test file — no source changes

Closes #635
Part of #632

### DIFF
--- a/packages/core/src/app/__tests__/bun-adapter.test.ts
+++ b/packages/core/src/app/__tests__/bun-adapter.test.ts
@@ -1,0 +1,136 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { createBunAdapter } from '../bun-adapter';
+
+// Mock the global Bun.serve that the adapter depends on
+const mockStop = vi.fn();
+const mockServer = {
+  port: 0,
+  hostname: '',
+  stop: mockStop,
+};
+const mockServe = vi.fn(() => mockServer);
+
+beforeEach(() => {
+  // Install the global Bun mock before each test
+  (globalThis as Record<string, unknown>).Bun = { serve: mockServe };
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+  mockServe.mockClear();
+  mockStop.mockClear();
+  // Clean up global
+  delete (globalThis as Record<string, unknown>).Bun;
+});
+
+describe('createBunAdapter', () => {
+  describe('listen', () => {
+    it('passes the port and handler to Bun.serve', async () => {
+      const adapter = createBunAdapter();
+      const handler = vi.fn();
+
+      mockServer.port = 4000;
+      mockServer.hostname = 'localhost';
+
+      await adapter.listen(4000, handler);
+
+      expect(mockServe).toHaveBeenCalledOnce();
+      const serveArg = mockServe.mock.calls[0][0];
+      expect(serveArg.port).toBe(4000);
+      expect(serveArg.fetch).toBe(handler);
+    });
+
+    it('passes hostname from options to Bun.serve', async () => {
+      const adapter = createBunAdapter();
+      const handler = vi.fn();
+
+      mockServer.port = 3000;
+      mockServer.hostname = '0.0.0.0';
+
+      await adapter.listen(3000, handler, { hostname: '0.0.0.0' });
+
+      const serveArg = mockServe.mock.calls[0][0];
+      expect(serveArg.hostname).toBe('0.0.0.0');
+    });
+
+    it('passes undefined hostname when no options are provided', async () => {
+      const adapter = createBunAdapter();
+      const handler = vi.fn();
+
+      mockServer.port = 3000;
+      mockServer.hostname = 'localhost';
+
+      await adapter.listen(3000, handler);
+
+      const serveArg = mockServe.mock.calls[0][0];
+      expect(serveArg.hostname).toBeUndefined();
+    });
+
+    it('passes undefined hostname when options exist but hostname is omitted', async () => {
+      const adapter = createBunAdapter();
+      const handler = vi.fn();
+
+      mockServer.port = 3000;
+      mockServer.hostname = 'localhost';
+
+      await adapter.listen(3000, handler, {});
+
+      const serveArg = mockServe.mock.calls[0][0];
+      expect(serveArg.hostname).toBeUndefined();
+    });
+
+    it('returns a ServerHandle with the port from Bun.serve', async () => {
+      const adapter = createBunAdapter();
+      const handler = vi.fn();
+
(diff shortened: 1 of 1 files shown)


viniciusdacal (commented):
## CHANGES REQUESTED

Overall: The tests are directionally correct — they test the adapter's transformation contract rather than just existence. However, there are specific issues that range from a meaningful behavioral gap to tests that are weaker than they appear.

### Finding 1: `handle.port` test reads from the same mutable object it configures — tautological

**File:** `bun-adapter.test.ts:87-94` ("returns a ServerHandle with the port from Bun.serve")

The test sets `mockServer.port = 8080`, then asserts `handle.port` is `8080`. But the adapter's source does `port: server.port` — it reads from the same `mockServer` object the test mutated. This test will pass even if the adapter hardcodes `port: 8080` or does `port: server.port + 0`, because the assertion value and the mock value are both written by the test.

The test at line 119 ("reflects the actual port Bun.serve binds to") partially addresses this by setting `mockServer.port = 54321` while requesting port `0`, proving the adapter reads from the server and not from the input argument. **That test is good.** But the test at line 87 is redundant with it and weaker — it could be removed without reducing mutation coverage.

**Impact:** Low. The port 0 test covers this. But the redundant test creates false confidence.

**Recommendation:** Remove the test at line 87 or combine it with the port 0 test. The port 0 scenario already proves the adapter reads `server.port`, not the input `port`.

### Finding 2: `handle.hostname` test has the same tautology problem — and no port-0-equivalent test saves it

**File:** `bun-adapter.test.ts:96-105` ("returns a ServerHandle with the hostname from Bun.serve")

The test sets `mockServer.hostname = '127.0.0.1'`, then asserts `handle.hostname === '127.0.0.1'`. The adapter does `hostname: server.hostname`. Since `mockServer` is the same mutable object, this will pass even if the adapter hardcodes `hostname: '127.0.0.1'`.

Unlike `port`, there is **no second test** that proves `hostname` is actually read from the server object (comparable to how the port 0 test proves `port` is read from the server). If the adapter did `hostname: 'localhost'` hardcoded, the test at line 96 would fail, but a mutation like `hostname: options?.hostname ?? 'localhost'` (reading from input instead of server) would survive undetected — there is no test where the server-returned hostname differs from what was passed in.

**Impact:** Medium. A mutant that reads hostname from the wrong source would survive.

**Recommendation:** Add a test where the adapter receives no explicit hostname option, but `mockServer.hostname` is set to something the adapter couldn't have derived from the input (e.g., `'0.0.0.0'` when no hostname was passed). This proves the handle reads the hostname from the server, not from `options`.

### Finding 3: Missing edge case — what happens when `Bun.serve` throws?

**File:** `bun-adapter.ts:13` — `Bun.serve()` is called synchronously; if it throws, the adapter's `listen()` promise rejects.

There is no test for this behavior. If `Bun.serve` throws (e.g., port already in use, invalid options), the adapter should propagate the error. This is a real consumer-facing scenario — developers will hit port conflicts.

**Impact:** Low-medium. The current implementation naturally propagates since it doesn't catch, but a future refactor that adds try/catch for logging could swallow the error. A test pins this behavior.

**Recommendation:** Add a test:
```typescript
it('propagates errors from Bun.serve', async () => {
  mockServe.mockImplementationOnce(() => { throw new Error('port in use'); });
  const adapter = createBunAdapter();
  await expect(adapter.listen(3000, vi.fn())).rejects.toThrow('port in use');
});
```

### Finding 4: `close()` test verifies `stop(true)` — this is good

**File:** `bun-adapter.test.ts:107-118`

This is the strongest test in the file. It verifies:
1. `stop` is not called prematurely (before `close()`)
2. `stop` is called exactly once after `close()`
3. `stop(true)` is called specifically (closing active connections)

The `true` argument matters — `stop(false)` would leave connections open. This test would catch that mutation. No notes.

### Finding 5: Hostname forwarding tests (lines 42-70) are solid

The three hostname tests cover: explicit hostname passed, no options provided, empty options provided. These verify the adapter correctly forwards `options?.hostname` to `Bun.serve`. The `undefined` assertions are important — they prove the adapter doesn't inject a default hostname.

### Finding 6: Handler forwarding test is correct but could be more explicit

**File:** `bun-adapter.test.ts:30-40`

The test checks `serveArg.fetch === handler` (reference equality). This is strong — it verifies the adapter passes the exact handler function, not a wrapper. If someone wraps the handler (for logging, error handling), this test would catch it. Good.

### Summary of Requested Changes

1. **Add a hostname-divergence test** (Finding 2) — proves `handle.hostname` reads from the server, not from options. This is the only real behavioral gap.
2. **Add an error propagation test** (Finding 3) — pins the behavior that `Bun.serve` errors propagate.
3. **Consider removing the redundant port test at line 87** (Finding 1) — optional, but it adds nothing over the port 0 test.

The `close()` test, handler forwarding test, and hostname option forwarding tests are all good behavioral tests that verify real contracts.

Issue #618, opened 2026-02-22
Migration integration: wire existing infra into createDbProvider + CLI


viniciusdacal:
## Description

Wire the existing migration infrastructure in `@vertz/db` into `createDbProvider()` for auto-migration in dev and CLI commands for production.

### Context
Per the migration design doc (`plans/design-schema-migrations.md`), most of the migration infrastructure **already exists** in `@vertz/db`:

| Component | Location | Status |
|-----------|----------|--------|
| `runner.ts` | `packages/db/src/migration/` | ✅ History table, checksums, drift detection |
| `snapshot.ts` | `packages/db/src/migration/` | ✅ `d.table()` → `SchemaSnapshot` |
| `differ.ts` | `packages/db/src/migration/` | ✅ Structural diff between snapshots |
| `sql-generator.ts` | `packages/db/src/migration/` | ✅ Dialect-specific SQL from diff |
| `push.ts` | `packages/db/src/cli/` | ✅ Direct apply |
| `migrate-dev.ts` | `packages/db/src/cli/` | ✅ Generate + apply + snapshot |
| `migrate-deploy.ts` | `packages/db/src/cli/` | ✅ Apply pending in order |
| `status.ts` | `packages/db/src/cli/` | ✅ Show pending |

**Whats missing:** Snapshot persistence (file-based), auto-migration in `createDbProvider()`, and wiring CLI commands.

#### 1. Wire migration infra into `createDbProvider()`
- Implement `SnapshotStorage` (file-based at `.vertz/schema-snapshot.json`)
- On init: load previous snapshot → diff against current schema → auto-apply in dev
- Dev/prod detection via `NODE_ENV` / `VERTZ_ENV`

```ts
const db = createDbProvider({
  dialect: \"sqlite\",
  schema: schema,
  connection: { path: \"./data.db\" },
  migrations: {
    autoApply: true, // default: true in dev, false in prod
    snapshotPath: \".vertz/schema-snapshot.json\",
  },
});
```

#### 2. Auto-migration in dev
- When schema changes are detected, auto-apply migrations
- Log warnings for destructive changes (drop table, drop column, alter type)
- Block destructive changes in production (require `--force`)

#### 3. CLI commands
```bash
vertz db push          # Direct apply for rapid prototyping (dev only)
vertz db migrate dev   # Generate migration file + apply locally
vertz db migrate deploy  # Apply pending migrations (CI/production)
vertz db status        # Show pending migrations
[...]

PR #639, opened 2026-02-23
feat(db): SnapshotStorage + migrations config for createDbProvider

viniciusdacal:
## Summary

Issue #618 - Migration subtask 1: Wire SnapshotStorage into createDbProvider

### Changes

1. **Created **
   -  - Load schema snapshot from JSON file (returns null if missing)
   -  - Save schema snapshot to JSON file

2. **Extended  config** in 
   - Added  option with:
     -  - default: 
     -  - default: 

### Note
Auto-migration is NOT implemented yet — this just accepts the config for future use.

## Testing
- TypeScript compiles successfully

### DIFF
--- a/packages/db/src/core/db-provider.ts
+++ b/packages/db/src/core/db-provider.ts
@@ -196,7 +196,7 @@ export function createDbProvider<TModels extends Record<string, ModelEntry>>(
         await autoMigrate({
           currentSchema,
           snapshotPath,
-          dialect: dialect as 'sqlite',
+          dialect: dialect as 'sqlite', // TODO: support postgres
           db: queryFn,
         });
       }


viniciusdacal (commented):
## 🔴 Adversarial Review: Migration Integration (PR #639)

Found several critical issues that need addressing before merge:

### 1. Destructive Changes Not Blocked
**File:** `auto-migrate.ts` lines 29-34, 87-96

The code detects destructive changes (`table_removed`, `column_removed`) and logs warnings, but **proceeds to execute them anyway**. This is dangerous - a dropped column means permanent data loss.

```typescript
// Only logs warning, doesn't stop migration
console.warn(`[auto-migrate] ⚠️  Warning: Destructive change detected...`);
```

**Fix needed:** Add option to halt on destructive changes or require explicit opt-in.

### 2. No Race Condition Protection
**File:** `auto-migrate.ts`

Multiple server instances starting simultaneously will all attempt to run migrations concurrently. No advisory locks or coordination mechanism.

**Fix needed:** Add DB-level locking or distributed lock.

### 3. Corrupted Snapshot Crashes App
**File:** `snapshot-storage.ts` lines 8-16

```typescript
} catch (err) {
  if ((err as NodeJS.ErrnoException).code === 'ENOENT') {
    return null;  // Only handles missing file
  }
  throw err;  // JSON parse errors CRASH the app
}
```

Corrupted JSON = application won't start. Should gracefully handle parse errors.

### 4. No Transaction Wrapping
**File:** `auto-migrate.ts` lines 104-112

If migration SQL fails mid-execution, partial changes are applied. No rollback.

**Fix needed:** Wrap in transaction or add idempotency checks.

### 5. Path Traversal Vulnerability
**File:** `db-provider.ts` line 189

```typescript
const snapshotPath = config.migrations?.snapshotPath ?? '.vertz/schema-snapshot.json';
```

User-controlled path could write outside intended directory (`../../etc/passwd`).

**Fix needed:** Validate path is within allowed directory.

### 6. Hardcoded SQLite Dialect
**File:** `db-provider.ts` line 194

```typescript
dialect: 'sqlite', // TODO: Detect from db driver
```

This will break for Postgres. Auto-detection needed before production use.

### 7. No Crash Recovery
If process crashes after applying SQL but before saving snapshot, changes re-apply on next restart. Could cause issues with non-idempotent operations.

### 8. Sensitive Data in Plaintext
Snapshot stores column metadata including `sensitive: true` flags in JSON. If snapshot file is leaked, sensitive column names are exposed.

---

**Recommendation:** Address items 1-5 before merge. Items 6-8 are important but can be deferred.

viniciusdacal (commented):
## Code Review: PR #639 - Migration integration

### Summary
Good implementation overall! The wiring of migration infrastructure into `createDbProvider` is solid and follows the design goals.

---

### ✅ What is Correct

**1. SnapshotStorage (load/save)**
- `loadSnapshot()` correctly returns `null` on first run (ENOENT)
- `saveSnapshot()` creates parent directories recursively
- Proper error handling for file I/O

**2. Auto-migrate logic**
- Diff computation works correctly
- Destructive changes (table_removed, column_removed) are properly warned
- SQL generation and application flow is correct
- Snapshot is saved after successful migration

**3. createDbProvider wiring**
- `shouldAutoMigrate()` correctly checks `autoApply` or defaults to `NODE_ENV !== production`
- `extractSchemaSnapshot()` properly converts models to snapshot format
- Integration in `onInit` is in the right place

**4. Tests**
- Auto-migrate integration tests cover the main scenarios (autoApply true/false, NODE_ENV detection)

---

### ⚠️ Issues and Suggestions

**1. Hardcoded dialect (Medium)**
Line 180 in db-provider.ts:
```
dialect: 'sqlite', // TODO: Detect from db driver
```
This will not work for PostgreSQL. Should either:
- Detect from the URL (e.g., url.startsWith("postgres"))
- Accept dialect in config
- Extract from the db instance

**2. Test coverage gaps**
- No unit tests for `snapshot-storage.ts` directly
- Tests use `:memory:` path which will not actually persist - the test verifies "no throw" but does not confirm migration actually ran
- `extractSchemaSnapshot()` has no dedicated unit tests

**3. Error handling**
- If `autoMigrate` fails, `onInit` fails completely. Consider whether migrations should be non-blocking or have a retry mechanism.

**4. Missing from exports**
- `loadSnapshot` and `saveSnapshot` are not exported from the migration index - users may need them for custom implementations

---

### 🎯 Recommendations

1. **Fix the dialect issue before merging** - this is a functional bug
2. Add at least one integration test that uses a real temp file to verify file persistence
3. Export `loadSnapshot`/`saveSnapshot` if intended for external use, or document that they are internal

---

Overall: **Approve with suggestions** - the core logic is sound and the dialect issue is a known TODO.

PR #642, opened 2026-02-23
test(core): address review feedback for bun-adapter tests

viniciusdacal:
## Summary

- Replaces the tautological port test (set `mockServer.port = 8080`, assert `handle.port === 8080`) with a **hostname divergence test**: no hostname option is passed, but `mockServer.hostname` returns `'0.0.0.0'` — this proves the adapter reads hostname from the server object, not from input options.
- Removes the redundant "returns a ServerHandle with the hostname from Bun.serve" test (its intent is now more precisely captured by the divergence test).
- Adds an **error propagation test**: when `Bun.serve` throws (e.g. port already in use), the error reaches the caller unchanged.

Addresses review feedback from PR #638.
Relates to #632.

### DIFF
--- a/packages/core/src/app/__tests__/bun-adapter.test.ts
+++ b/packages/core/src/app/__tests__/bun-adapter.test.ts
@@ -79,28 +79,32 @@ describe('createBunAdapter', () => {
       expect(serveArg.hostname).toBeUndefined();
     });
 
-    it('returns a ServerHandle with the port from Bun.serve', async () => {
-      const adapter = createBunAdapter();
-      const handler = vi.fn();
-
-      mockServer.port = 8080;
-      mockServer.hostname = '127.0.0.1';
-
-      const handle = await adapter.listen(8080, handler);
-
-      expect(handle.port).toBe(8080);
-    });
-
-    it('returns a ServerHandle with the hostname from Bun.serve', async () => {
+    it('returns a ServerHandle whose hostname comes from the server object, not the input options', async () => {
+      // When no hostname option is passed, the adapter must read hostname from server.hostname,
+      // not derive it from the input. Here Bun binds to '192.168.1.100' even though no hostname
+      // was given — the handle must reflect what the server actually reports. Using a clearly
+      // non-default hostname ensures a broken implementation like `options?.hostname ?? '0.0.0.0'`
+      // would fail this test.
       const adapter = createBunAdapter();
       const handler = vi.fn();
 
       mockServer.port = 3000;
-      mockServer.hostname = '127.0.0.1';
+      mockServer.hostname = '192.168.1.100'; // server returns this even though no hostname option was provided
 
-      const handle = await adapter.listen(3000, handler);
+      const handle = await adapter.listen(3000, handler); // no hostname in options
 
-      expect(handle.hostname).toBe('127.0.0.1');
+      expect(handle.hostname).toBe('192.168.1.100');
+    });
+
+    it('propagates errors thrown by Bun.serve to the caller', async () => {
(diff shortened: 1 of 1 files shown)


viniciusdacal (commented):
## Adversarial Review: PR #642

**Verdict: CHANGES REQUESTED**

The direction is right — replacing the tautological port test and adding error propagation coverage are good moves. But the hostname divergence test has a specific weakness that a mutation would survive, which is the core goal of this PR.

---

### Issues Found

**1. Hostname divergence test uses `'0.0.0.0'` — a value a broken impl could hardcode (critical)**

The new test at line 89 sets `mockServer.hostname = '0.0.0.0'` and asserts `handle.hostname === '0.0.0.0'`. The intent is to prove the adapter reads `server.hostname`, not `options?.hostname`.

But the divergence only covers the mutation `return { hostname: options?.hostname }` (which returns `undefined`). It does NOT cover a more realistic broken implementation:

```ts
// broken: hardcodes the common Bun default instead of reading from server
return {
  hostname: options?.hostname ?? '0.0.0.0',
  ...
};
```

This broken implementation would pass the test because `options?.hostname` is `undefined`, so it falls back to the hardcoded string `'0.0.0.0'` — which happens to equal `mockServer.hostname`. The test cannot distinguish "read from server" from "hardcode the same string the server happens to return."

The fix is to use a hostname value that could not plausibly be derived from any input or common default:

```ts
mockServer.hostname = '192.168.1.100'; // or 'bun-test-host', or '10.0.0.42'
```

With a distinctive value, any impl that doesn't read `server.hostname` would return something other than `'192.168.1.100'`, making the mutation detectable.

The existing `'reflects the actual port Bun.serve binds to'` test gets this right — it uses port `54321` for the mock (clearly synthetic) against input `0`. This test should follow the same pattern.

---

**2. The deleted port test is justified — but the asymmetry is worth noting**

The PR removes the tautological port test (set port 8080, listen on 8080, assert 8080) without a port-specific replacement. Port divergence coverage comes entirely from `'reflects the actual port Bun.serve binds to'` (54321 vs 0), which does provide real divergence. Acceptable as-is, but worth adding a comment so future maintainers don't add back a tautological port test thinking it fills a gap.

---

**3. Error propagation test: `toThrow` matches by substring, not exact error**

```ts
await expect(adapter.listen(3000, handler)).rejects.toThrow('address already in use');
```

This passes even if the adapter wraps the error (`throw new Error('wrapped: address already in use')`). Tighter assertion:

```ts
await expect(adapter.listen(3000, handler)).rejects.toThrow(
  new Error('address already in use')
);
```

`toThrow(new Error('...'))` checks that the thrown value is an `Error` instance with that exact message. Minor improvement but closes the wrapping mutation.

---

### What's Good

- **Error propagation test is a real improvement.** This behavior was entirely untested before. `mockImplementationOnce` is correct — consumed on first call, and `mockClear()` in `afterEach` handles any unconsumed implementations. The test would catch swallowed errors, wrapped re-throws, or default return values.

- **Removing the old tautological port test is the right call.** `'reflects the actual port Bun.serve binds to'` (54321 vs input 0) already provides divergence proof for port. No duplicate needed.

- **Test naming is now meaningful.** The new names describe the invariant being tested, not just what the test does mechanically.

---

### Required Fix

Change line 90 in the PR branch from:

```ts
mockServer.hostname = '0.0.0.0'; // server returns this even though no hostname option was provided
```

To:

```ts
mockServer.hostname = '192.168.1.100'; // synthetic value — proves adapter reads from server, not from input or any Bun default
```

Update the surrounding comment accordingly. This is a one-line change that makes the divergence genuinely mutation-proof.

Issue #643, opened 2026-02-23
Extract SQLite/D1 adapters from entity-todo into @vertz/db


viniciusdacal:
## Description

The entity-todo example currently has full CRUD implementations in `db.ts` (SQLite) and `db-d1.ts` (D1). These should be generic adapters in `@vertz/db` so users don't need boilerplate.

## Goal

After this, entity-todo should just have:
```ts
const db = createDbProvider({
  dialect: 'sqlite',
  schema: schema,
  migrations: { autoApply: true },
});
```

No manual SQL, no CRUD implementations — the framework generates it from the schema.

## Tasks

1. Move SQLite adapter logic into `@vertz/db` as a generic adapter
2. Move D1 adapter logic into `@vertz/db` as a generic adapter
3. Generate CRUD operations from entity schema (create, list, get, update, delete)
4. Update entity-todo to use the generic adapters
5. Remove boilerplate `db.ts` and `db-d1.ts` from example

## Acceptance Criteria

- [ ] `createDbProvider()` returns an object with CRUD methods generated from schema
- [ ] SQLite adapter in `@vertz/db` works for local dev
- [ ] D1 adapter in `@vertz/db` works for Cloudflare
- [ ] entity-todo has no manual SQL or CRUD boilerplate
- [ ] SQL injection protection built into the adapters

## Labels

enhancement, P1

PR #646, opened 2026-02-23
feat(db): Extract SQLite/D1 adapters into @vertz/db (#643)

viniciusdacal:
This PR extracts the SQLite and D1 adapters from entity-todo into @vertz/db, following the goal of Issue #643.

## Changes

1. **Created generic SQLite adapter** (`packages/db/src/adapters/sqlite-adapter.ts`):
   - Takes schema + connection config
   - Implements: create, list, get, update, delete
   - Uses schema metadata to generate SQL — no manual SQL

2. **Created generic D1 adapter** (`packages/db/src/adapters/d1-adapter.ts`):
   - Same interface as SQLite adapter
   - Uses D1 binding instead of bun:sqlite

3. **Added `createDbProvider()` factory** (`packages/db/src/adapters/index.ts`):
   - Unified API for both SQLite and D1 dialects
   - Handles dialect-specific configuration

4. **Updated @vertz/db exports** (`packages/db/src/index.ts`)

5. **Updated entity-todo** to use the new generic adapters:
   ```ts
   import { createDbProvider } from @vertz/db;
   import { todosTable } from './schema';
   
   export const db = createDbProvider({
     dialect: 'sqlite',
     schema: todosTable,
     migrations: { autoApply: true },
   });
   ```

## Testing

- API tests pass (6/6)
- Dev server starts correctly
- Build succeeds

### DIFF
--- a/packages/db/src/adapters/__tests__/sqlite-adapter.test.ts
+++ b/packages/db/src/adapters/__tests__/sqlite-adapter.test.ts
@@ -38,7 +38,7 @@ describe('SqliteAdapter', () => {
     driver = createSqliteDriver(':memory:');
 
     // Create the adapter with auto-apply migrations
-    adapter = createSqliteAdapter<UsersSchema>({
+    adapter = await createSqliteAdapter<UsersSchema>({
       schema: usersTable,
       dbPath: ':memory:',
       migrations: { autoApply: true },


viniciusdacal:
## 🔴 Adversarial Review — Critical Issues Found

I've reviewed this PR with a focus on SQL injection, API inconsistencies, edge cases, and better-sqlite3 compatibility. **Found multiple critical security issues that need to be addressed before merge.**

---

## 🚨 CRITICAL: SQL Injection Vulnerabilities

### 1. **Table Name Not Escaped (CRITICAL)**

Both `sqlite-adapter.ts` and `d1-adapter.ts` directly interpolate `schema._name` into SQL without validation or escaping:

```typescript
// sqlite-adapter.ts:217
const rows = await this.driver.query<Record<string, unknown>>(
  `SELECT * FROM ${this.tableName} WHERE id = ?`,  // ⚠️ tableName not escaped
  [id],
);
```

Same issue in `INSERT`, `UPDATE`, `DELETE`, `CREATE TABLE`, and `CREATE INDEX` statements.

**Attack vector:** If a schema is ever constructed with user-influenced table name:
```typescript
const maliciousName = "users; DROP TABLE users; --";
const table = d.table(maliciousName, {...});
```

**Fix:** Escape identifiers or validate table names against a strict pattern (alphanumeric + underscore).

---

### 2. **Column Names Not Escaped (CRITICAL)**

Column names are interpolated directly without escaping:

```typescript
// sqlite-adapter.ts:278
whereClauses.push(`${key} = ?`);  // ⚠️ key not escaped
```

While there's whitelist validation for `where` clauses in `list()`, the following are NOT validated:

- `generateCreateTableSql()` — column definitions
- `generateIndexSql()` — index columns  
- `create()` — column names in INSERT
- `update()` — column names in UPDATE

**Example issue in index generation:**
```typescript
// sqlite-adapter.ts:122
sqls.push(`CREATE INDEX IF NOT EXISTS ${indexName} ON ${tableName} (${index.columns.join(', ')})`);
```

If `index.columns` contains `["id); DROP TABLE users; --"]`, this would execute the malicious SQL.

---

### 3. **CHECK Constraints Not Escaped (CRITICAL)**

```typescript
// sqlite-adapter.ts:80
if (meta.check) {
  colDef += ` CHECK (${meta.check})`;  // ⚠️ meta.check directly interpolated
}
```

If `meta.check` contains malicious SQL, it will be executed.

---

### 4. **Index Names Not Escaped**

```typescript
// sqlite-adapter.ts:117
const indexName = index.name || `idx_${tableName}_${index.columns.join('_')}`;
sqls.push(`CREATE ${unique}INDEX IF NOT EXISTS ${indexName} ...`);
```

---

## ⚠️ HIGH: API Inconsistencies

### 5. **Boolean WHERE Clause Handling Differs Between Adapters**

**SQLite adapter** (line 265-267):
```typescript
// Converts boolean to integer for WHERE clause
const colMeta = this.schema._columns[key]?._meta as ColumnMetadata | undefined;
const convertedValue = colMeta?.sqlType === 'boolean' ? (value ? 1 : 0) : value;
```

**D1 adapter** (line 269-271):
```typescript
// Does NOT convert boolean — just passes the value through
countParams.push(value);
```

This means `list({ where: { active: true } })` works in SQLite but **fails silently in D1** because D1 stores booleans as INTEGER.

---

### 6. **Hardcoded Primary Key Column**

Both adapters assume the primary key is always named `"id"`:

```typescript
// sqlite-adapter.ts:217
`SELECT * FROM ${this.tableName} WHERE id = ?`
```

But a schema could have a different primary key:
```typescript
d.table('users', {
  userId: d.uuid().primary(),  // Primary key is "userId", not "id"
  name: d.text(),
});
```

This would cause `get()`, `update()`, and `delete()` to fail silently.

---

### 7. **Missing Limit Bounds**

```typescript
// sqlite-adapter.ts:287
const limit = options?.limit ?? 20;
sql += ' LIMIT ?';
params.push(limit);
```

No validation:
- Negative limits are accepted
- Extremely large limits could cause memory issues
- No maximum bound enforced

---

## ⚠️ MEDIUM: Edge Cases & Better-sqlite3 Compatibility

### 8. **Mutating Input Data Object**

```typescript
// sqlite-adapter.ts:310
if (meta.primary && !data[colName]) {
  data[colName] = crypto.randomUUID();  // ⚠️ Mutates caller's object
}
```

This violates the principle of immutability and can cause unexpected behavior:
```typescript
const input = { name: "Test" };
await adapter.create(input);
console.log(input.id);  // Surprise! Now has an id property
```

---

### 9. **Error Information Loss**

Original errors are swallowed and replaced with generic messages:

```typescript
// sqlite-adapter.ts:226
} catch (error) {
  throw new Error(`Failed to retrieve record: resource may be unavailable`);
}
```

This makes debugging extremely difficult. The original error should be chained or logged.

---

### 10. **Missing `cuid()` Implementation**

```typescript
// sqlite-adapter.ts:312
} else if (meta.generate === 'cuid') {
  data[colName] = crypto.randomUUID(); // TODO: use actual cuid
}
```

If someone uses `d.uuid().primary({ generate: 'cuid' })`, they get a UUID instead of a CUID. This is a silent bug.

---

### 11. **Potential Race Condition in Driver Fallback**

```typescript
// sqlite-adapter.ts:161
try {
  const { Database } = require('bun:sqlite');
  db = new Database(dbPath);
} catch {
  try {
    const Database = require('better-sqlite3');
    db = new Database(dbPath);
  } catch (error) {
```

If `bun:sqlite` exists but fails to open the database (e.g., file locked), it falls through to `better-sqlite3`, which might also fail but with a misleading error message.

---

### 12. **Missing `exec()` Method in Interface**

The `SqliteDatabase` interface declares `exec()` but `better-sqlite3`'s `exec()` behaves differently than `bun:sqlite`'s. The PRAGMA call:
```typescript
db.exec('PRAGMA journal_mode = WAL');
```
works in bun:sqlite but `better-sqlite3`'s `exec()` doesn't return anything useful for PRAGMA statements.

---

## 📋 Recommended Fixes

1. **Add identifier escaping** — Create an `escapeIdentifier()` function and use it for all table/column/index names
2. **Validate table names** — Only allow `[a-zA-Z_][a-zA-Z0-9_]*`
3. **Validate column names** — Same pattern
4. **Fix D1 boolean WHERE clause** — Port the conversion logic from SQLite adapter
5. **Dynamic primary key detection** — Use schema metadata to find PK column
6. **Add limit bounds** — Clamp to [1, 1000] or similar
7. **Don't mutate input** — Create a copy before modifying
8. **Preserve error context** — Use `error.cause` or wrap with original error
9. **Implement or remove cuid** — Either implement proper CUID or throw an error
10. **Add tests for edge cases** — Custom PK names, empty data, negative limits

---

## Summary

| Severity | Count |
|----------|-------|
| Critical (SQL Injection) | 4 |
| High (API Inconsistency) | 3 |
| Medium (Edge Cases) | 5 |

**Recommendation:** Block merge until SQL injection issues are addressed. The code is well-structured but needs security hardening before production use.

PR #662, opened 2026-02-23
fix(ui-server): resolve SSR rendering artifacts

### LINKED ISSUE #661: fix(ui-server): SSR rendering artifacts — styles, conditionals, object serialization
## Context

PR #659 fixed SSR crashes (JSX runtime swap, hardened isSSR, ThemeProvider). The server now renders HTML without errors, but the output has rendering artifacts that make the page unusable.

### 1. Styles not applied
The SSR HTML has no `<style>` tags with component CSS. The `globalStyles.css` and component-level styles (`css()`, `variants()`) aren't being injected into the HTML `<head>`.

**Root cause:** `renderToHTML()` in `@vertz/ui-server` collects theme CSS and `options.styles`, but component-level CSS from `css()`/`variants()` calls are injected into the DOM at runtime (client-side only). During SSR, these styles are generated but never collected into the HTML output.

**Fix direction:** The SSR renderer needs to collect all CSS generated during the render pass and inject it as `<style>` tags in the HTML head. Options:
- A global CSS collector that `css()`/`variants()` write to during SSR
- Or extract critical CSS from the component tree after rendering

### 2. `[object Object]` in output
Some VNodes are being serialized as `[object Object]` instead of rendered to HTML. This happens when the VNode serializer encounters a node type it doesn't recognize.

**Root cause:** The JSX runtime swap produces VNodes from `@vertz/ui-server/jsx-runtime`, but some component return values (e.g., ThemeProvider's VNode, conditional branches) don't match the expected `VNode` interface exactly, or the serializer calls `.toString()` on objects.

**Fix direction:** Check `renderPage()`'s serialization logic — it needs to handle:
- Plain objects with `{ tag, attrs, children }` shape (ThemeProvider SSR return)
- Nested arrays of children
- null/undefined/boolean children (should be skipped)

### 3. `<!-- conditional -->` comments visible
Conditional rendering (`{condition && <Component />}`) produces comment markers that leak into the HTML output.

**Root cause:** The `@vertz/ui` runtime uses comment nodes as placeholders for conditional branches. During SSR with the JSX swap, these become string literals instead of being stripped.

**Fix direction:** The server JSX runtime or the HTML serializer should:
- Strip comment-like strings from children
- Or handle conditional rendering differently in VNode mode (skip falsy branches entirely)

### 4. POST /api/todos returns 500
Separate from rendering — the entity CRUD handler crashes on create. Likely a missing field or schema validation issue in the entity handler.

## Files to investigate
- `packages/ui-server/src/render-page.ts` — HTML serialization
- `packages/ui-server/src/render-to-html.ts` — renderToHTML orchestration
- `packages/ui-server/src/jsx-runtime/index.ts` — server JSX runtime
- `packages/ui/src/css/css.ts` — css() runtime injection
- `packages/ui/src/css/variants.ts` — variants() runtime injection
- `examples/entity-todo/src/entry-server.ts` — SSR entry
- `examples/entity-todo/src/pages/todo-list.tsx` — conditionals that produce artifacts

## Acceptance criteria
- [ ] SSR HTML includes component styles (not just theme CSS)
- [ ] No `[object Object]` in rendered HTML
- [ ] No `<!-- conditional -->` artifacts in HTML
- [ ] Page looks styled when loading (before client JS runs)
- [ ] POST /api/todos works (create a todo)

Refs: #659, #660

viniciusdacal:
## Summary

Fixes 4 SSR rendering artifacts reported in #661:

- **Comment nodes visible in HTML** — `createComment()` now returns empty `SSRTextNode` instead of literal `<!-- text -->` markup that was being escaped and rendered as visible text
- **`[object Object]` in output** — server JSX runtime now unwraps signal-like objects via `peek()` in both children and attributes (duck-typed check for `.peek` method)
- **Component CSS not collected** — `renderToHTML()` now calls `app()` first (triggering `css()`/`injectCSS()` calls), then collects CSS from `document.head.children` before building style tags
- **POST /api/todos 500** — dev-server `apiMiddleware` now wraps the async handler in try/catch with proper JSON error response instead of unhandled promise rejection

All fixes follow strict TDD with dedicated test cases (143 ui-server tests + 7 entity-todo API tests passing).

Closes #661

## Test plan

- [x] `bun test packages/ui-server` — all 143 tests pass (existing + new)
- [x] `bun test examples/entity-todo` — all 7 tests pass (existing + new)
- [x] `tsc --noEmit --project packages/ui-server/tsconfig.json` — clean
- [x] `bunx biome check packages/ui-server/src/ examples/entity-todo/src/` — clean
- [ ] Manual: start dev server (`bun run dev` in entity-todo), verify no `<!-- conditional -->` or `[object Object]` in SSR HTML
- [ ] Manual: `curl -X POST localhost:3000/api/todos` with empty body returns JSON error, not 500 hang

### DIFF
--- a/packages/ui-server/src/render-to-html.ts
+++ b/packages/ui-server/src/render-to-html.ts
@@ -1,7 +1,7 @@
 import { compileTheme, type Theme } from '@vertz/ui';
 import { installDomShim, removeDomShim, SSRElement } from './dom-shim';
 import { renderPage } from './render-page';
-import { ssrStorage, getSSRUrl } from './ssr-context';
+import { ssrStorage } from './ssr-context';
 import type { VNode } from './types';
 
 export interface RenderToHTMLOptions<AppFn extends () => VNode> {


viniciusdacal (commented):
Review complete. See attached file for details.

viniciusdacal:
Test comment on issue

PR #663, opened 2026-02-23
fix(ui): SSR DOM compiler path — VNode conversion and reactive evaluation

### LINKED ISSUE #661: fix(ui-server): SSR rendering artifacts — styles, conditionals, object serialization

viniciusdacal:
## Summary

- **Fix `[object Object]` in SSR output**: `__insert` now detects VNode-like objects (from Vite's server JSX runtime) and converts them to DOM nodes instead of stringifying them
- **Fix empty reactive content during SSR**: All reactive DOM primitives (`__text`, `__child`, `__conditional`, `__attr`, `__list`) now evaluate content synchronously during SSR instead of wrapping in `effect()` (which is a no-op during SSR)
- **Export `isSSR()`**: Made the SSR detection utility available to DOM primitives, with fallback from function hook (`__VERTZ_IS_SSR__`) to boolean flag (`__VERTZ_SSR__`)

### Root Cause

PR #662 fixed SSR artifacts in the **JSX runtime path**, but apps using `@vertz/ui-compiler` (DOM operations path) had two remaining bugs:

1. Non-reactive expressions containing JSX produce VNode objects → `String(vnode)` → `[object Object]`
2. All reactive primitives wrap logic in `effect()` → SSR no-op → zero content

### Changes

| File | Fix |
|------|-----|
| `packages/ui/src/runtime/signal.ts` | Export `isSSR()` with function hook + boolean fallback |
| `packages/ui/src/dom/element.ts` | `isVNode`/`vnodeToDOM`/`unwrapSignal` helpers + SSR paths for `__insert`, `__text`, `__child` |
| `packages/ui/src/dom/conditional.ts` | `ssrConditional` — synchronous evaluation |
| `packages/ui/src/dom/attributes.ts` | `__attr` SSR path — synchronous attribute setting |
| `packages/ui/src/dom/list.ts` | `__list` SSR path — synchronous item rendering |
| `packages/ui/src/dom/__tests__/ssr-rendering.test.ts` | 36 tests covering all SSR fixes |
| `packages/ui-server/src/render-to-html.ts` | Activate SSR flag via `ssrStorage.run()` |

## Test plan

- [x] 36 new SSR rendering tests pass (all DOM primitives in SSR context)
- [x] 1038 existing `@vertz/ui` tests pass (no regressions)
- [x] 143 existing `@vertz/ui-server` tests pass (no regressions)
- [x] TypeScript strict mode clean on both packages
- [x] Biome lint clean
- [x] Pre-push quality gates pass (60/60 tasks)
- [ ] E2E: Start entity-todo dev server, view source — HTML contains rendered content, no `[object Object]`, no empty reactive sections

Closes #661

### DIFF
--- a/examples/entity-todo/src/api/client.ts
+++ b/examples/entity-todo/src/api/client.ts
@@ -6,11 +6,17 @@
  * for compile-time exhaustiveness checking via matchError.
  */
 
-import { FetchClient, type FetchClientConfig, type Result, type FetchErrorType } from '@vertz/fetch';
+import {
+  FetchClient,
+  type FetchClientConfig,
+  type FetchErrorType,
+  type Result,
+} from '@vertz/fetch';
 import { createTodosSdk } from '../generated/entities/todos';
 
-// Base URL for the API (defaults to local dev server)
-const API_BASE = process.env.API_BASE_URL || 'http://localhost:3000/api';
+// Base URL for the API — derived from current origin in browser, falls back for SSR/Node
+const API_BASE =
+  typeof window !== 'undefined' ? `${window.location.origin}/api` : 'http://localhost:3000/api';
 
 /**
  * Create an SDK client instance.
@@ -49,7 +55,9 @@ export type UpdateTodoInput = Partial<Pick<Todo, 'title' | 'completed'>>;
  *
  * @returns Result containing array of todos or FetchError
  */
-export async function fetchTodos(): Promise<Result<{ todos: Todo[]; total: number }, FetchErrorType>> {
+export async function fetchTodos(): Promise<
+  Result<{ todos: Todo[]; total: number }, FetchErrorType>
+> {
   const result = await sdk.list();
 
   if (result.ok) {
@@ -136,7 +144,9 @@ export async function updateTodo(
 /**
  * Delete a todo.
  */
-export async function deleteTodo(id: string): Promise<Result<{ success: boolean }, FetchErrorType>> {
+export async function deleteTodo(
+  id: string,
+): Promise<Result<{ success: boolean }, FetchErrorType>> {
   const result = await sdk.delete(id);
 
   if (result.ok) {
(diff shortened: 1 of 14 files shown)


viniciusdacal:
Superseded by #662 which was already merged to main.

Issue #660, opened 2026-02-23
feat(ui): Make effect() internal-only + SSR-safe


viniciusdacal:
## Problem

`effect()` is currently exported as a public API from `@vertz/ui`. This creates two problems:

### 1. SSR Breakage
`effect()` executes synchronously during SSR where DOM APIs don't exist (`document`, `Node`, etc.). This causes `ReferenceError: Node is not defined` and `document.querySelector is not a function` errors. See entity-todo demo crash.

### 2. Footgun API
Raw `effect()` is the same trap as React's `useEffect` — it's powerful but dangerous. Users write effects that:
- Run at wrong times
- Leak subscriptions
- Create cascading updates
- Mix side effects with rendering

This leads to exactly the kind of spaghetti code we're trying to eliminate. We don't want a `useEffect` equivalent polluting the ecosystem.

### Make `effect()` internal-only
- Remove `effect` from `@vertz/ui` public exports
- Keep it as an internal primitive used by `query()`, `computed()`, `watch()`, etc.
- Framework code controls where and how effects run

### Make internal `effect()` SSR-safe
- Detect SSR context (no `document`, no `window`, or explicit SSR flag)
- Skip effect execution during SSR — effects are side-effectful by nature, they belong on the client
- `computed()` can still work during SSR (pure derivation, no DOM access)

### User-facing alternatives
Users should use higher-level, constrained patterns instead:

| Instead of | Use | Why |
|---|---|---|
| `effect(() => fetch(...))` | `query()` | Handles loading, error, caching, refetch |
| `effect(() => derived)` | `computed()` | Pure derivation, SSR-safe |
| `effect(() => setup)` | `onMount()` | Runs once on client mount |
| `effect(() => watch signal)` | `watch(signal, callback)` (TBD) | Constrained reactivity |

### Migration for entity-todo
`todo-list.tsx` currently uses `effect()` to bridge query results to local variables. This should use `computed()` or the query's built-in reactive properties directly.

## Scope

- [ ] Remove `effect` from `@vertz/ui` public exports
- [ ] Add SSR context detection to internal `effect()` (skip execution in SSR)
- [ ] Audit all usages of `effect()` in examples — replace with proper patterns
- [ ] Design `watch()` API if needed (constrained signal observation)
- [ ] Update entity-todo to not use `effect()` directly
- [ ] Document the "no public effect" decision and alternatives

## Principles

- **If it confuses an LLM, it confuses a junior developer** — effect() confuses both
[...]

PR #665, opened 2026-02-23
docs: universal rendering model design doc

### LINKED ISSUE #664: feat(ui): Universal rendering model — one pipeline, every target
## Summary

Replace the three-path rendering model (CSR / SSR / Hydration) with a **single universal rendering pipeline**. Components produce the same output everywhere — the only thing that changes is the output adapter (serialize to HTML, attach to live DOM, map to native widgets).

## Problem

After fixing SSR DOM compiler bugs in PR #662 and PR #663, it's clear the root cause of recurring SSR issues is the three divergent code paths. Every DOM primitive (`__text`, `__child`, `__conditional`, `__attr`, `__list`, `__insert`) has three branches:

1. **CSR** — `effect()` runs, DOM is live
2. **SSR** — `effect()` is a no-op, DOM is empty (the bug class we keep patching)
3. **Hydration** — client claims existing DOM, re-attaches effects

This means every new feature multiplies across three paths, and `effect()` being a no-op during SSR is a permanent source of bugs.

### Universal render pipeline

Every component, everywhere, does the same thing:

1. **Synchronous render** — component function runs, produces a node tree, effects run once to populate initial state
2. **Output adapter** determines what happens next:

| Target | After render |
|--------|-------------|
| Browser (first load) | Serialize → HTML → hydrate → effects track reactivity |
| Browser (navigation) | Tree IS the live DOM → effects track reactivity |
| Server | Serialize → HTML string → discard tree |
| Tauri / Electron-Bun | Same as browser — real webview |
| Custom Bun native runtime | Node tree maps to native widgets → effects update them |

### Effect behavior

| Environment | Callback runs? | Subscriptions tracked? |
|------------|---------------|----------------------|
| Browser | Yes | Yes (live DOM needs reactivity) |
| SSR | Yes (populates initial state) | No (tree is serialized and discarded) |
| Native app | Yes | Yes (native widgets need reactivity) |

During SSR, effects run their callback synchronously to populate DOM content, but `setSubscriber()` is not called — no signal tracking, no wasted memory. This is simpler than the current no-op and produces correct output.

### `effect()` is internal-only

`effect()` is NOT a public API. Users use:
- `watch()` — react to signal changes (side effects)
- `onMount()` / `onCleanup()` — lifecycle hooks
- `computed()` / `const` declarations — derived state
- `query()` — async data fetching

This is already the convention (`.claude/rules/ui-components.md`) but becomes a hard architectural boundary.

### SSR data loading with threshold

When rendering on the server, if `query()` data resolves within a configurable threshold (e.g., 100ms), include the resolved data in SSR output instead of shipping a loading state. First paint with real content instead of spinners.

### Atomic component streaming

Stream individual components during SSR for progressive rendering. Foundation for future **live deployment streaming**: deploy a component change → server re-renders just that component → streams new HTML to connected clients → component swaps in place (only if user is not interacting with it). No page reload.

### Native app support

The SSR DOM shim (`SSRElement`, `SSRTextNode`) is already a prototype of a native adapter. Formalize this as an adapter interface:
- `document.createElement('div')` → native container
- `el.setAttribute(...)` → native styling
- `el.appendChild(child)` → native layout
- Effects update native widgets reactively

Enables: Tauri, Electron-Bun, custom Bun runtime, TUI (already exists).

## What gets deleted

- `isSSR()` checks in all DOM primitives
- SSR-specific branches in `__text`, `__child`, `__insert`, `__conditional`, `__attr`, `__list`
- `effect()` no-op guard in `signal.ts`
- VNode duck-typing (`isVNode`, `vnodeToDOM`, `unwrapSignal`) from PR #663
- `ssr-rendering.test.ts` (36 tests for code paths that won't exist)

## What stays the same

- Component authoring model (unchanged for users)
- `watch()`, `onMount()`, `computed()`, `query()`, `form()` public APIs
- Compiler output (`__text`, `__child`, etc. — same calls, simpler implementation)
- Hydration (still needed for browser first-load)

## Implementation phases

1. **Universal effect model** — remove SSR no-op, make effects run-once during SSR without tracking
2. **Simplify DOM primitives** — remove three-path branching, single code path
3. **SSR data threshold** — `query()` waits for fast data during SSR
4. **Component streaming** — progressive SSR rendering
5. **Output adapter interface** — formalize for native targets
6. **Live deploy streaming** — component identity, interaction detection, incremental builds (future)

## Open questions

1. Threshold default for SSR data loading — 50ms? 100ms? Configurable per-query?
2. Streaming granularity — per-component or per-Suspense-boundary?
3. Live deploy streaming — interaction detection heuristic (focus? mouse? active form state?)
4. Adapter interface — minimal surface a native adapter must implement?
5. Effect cleanup during SSR — run cleanup functions? (Probably not — tree is discarded)

## Manifesto alignment

- **One way to do things** — one rendering model, not three
- **Explicit over implicit** — no hidden environment-dependent behavior
- **Production-ready by default** — SSR with data loading out of the box
- **Convention over configuration** — SSR is the default, not an opt-in

## References

- PR #662: Fixed SSR artifacts in JSX runtime path
- PR #663: Fixed SSR artifacts in DOM compiler path (the 8 fixes that prompted this discussion)
- Design doc: `plans/universal-rendering-model.md`

viniciusdacal:
## Summary

Design doc for the universal rendering model proposed in #664, incorporating the lifecycle API simplification from #660.

- **Universal effect model**: Replace the three-path rendering (CSR/SSR/Hydration) with a single pipeline where `effect()` runs once during SSR without signal tracking, instead of no-opping
- **Lifecycle API simplification** (#660): Make `effect()`, `signal()`, `watch()`, `onCleanup()` internal-only. `onMount()` returns a cleanup function.
- **Automatic query field selection**: Documents why internal-only primitives are critical for the `FieldAccessAnalyzer`/`CrossComponentAnalyzer` to trace data flow and compile VertzQL queries with automatic field selection (GraphQL-like, but at build time)
- **Bun migration path**: Future phases to replace Vite with Bun's native dev server, HTML entrypoints, and `onBeforeParse` native plugin API for compiler transforms

## What this doc covers

| Section | Content |
|---------|---------|
| Problem | Three divergent code paths cause recurring SSR bugs |
| Proposal | Universal render pipeline + lifecycle API simplification |
| API Surface | `onMount` return-cleanup pattern, complete public API |
| Query field selection | Compiler traces field access across components for VertzQL |
| Implementation phases | 6 phases from universal effects → component streaming |
| Bun migration | Future path from Vite → Bun dev server → native compiler |
| Unknowns | Effect cleanup in SSR, watch() disposition, thresholds |

## Review requested on

1. **Universal effect model** — Is "run once, no tracking" the right SSR behavior?
2. **Lifecycle API changes** — Is removing `effect`, `signal`, `watch`, `onCleanup` from public exports the right call?
3. **`onMount` return cleanup** — Better pattern than standalone `onCleanup()`?
4. **Bun migration timeline** — Should we start Phase A (replace Vite) sooner?
5. **Query field selection** — Any concerns about the compiler needing full control over data flow?

Closes #664

### DIFF
--- a/plans/universal-rendering-model.md
+++ b/plans/universal-rendering-model.md
@@ -205,22 +205,66 @@ The returned function runs when the component is disposed. This mirrors the patt
 
 #### `onCleanup()` becomes internal-only
 
-With `onMount` returning the cleanup, standalone `onCleanup()` is no longer needed in user code. It stays internal for `watch()` and framework DOM helpers.
+With `onMount` returning the cleanup, standalone `onCleanup()` is no longer needed in user code. It stays internal for DOM primitives.
 
-#### `watch()` — internal-only, SSR-safe
+#### `watch()` — removed entirely
 
-**Decision (resolved from Unknown #2):** Keep the implementation, make it internal-only, skip during SSR.
+**Decision (resolved from Unknown #2):** Remove `watch()`. It has no legitimate callers.
 
-`watch()` uses `lifecycleEffect()` internally, which means it's a no-op during SSR. This is correct:
-- `RouterView` uses `watch()` to swap pages on navigation → no navigation happens during SSR
-- Theme change listeners → no theme changes during SSR
-- localStorage sync → no localStorage on the server
+The only production caller is `RouterView`, which uses `watch()` for imperative DOM manipulation — `container.innerHTML = ''` + `appendChild()`. This is an anti-pattern: the framework teaches developers to use declarative JSX, but `RouterView` itself builds DOM imperatively. It doesn't eat its own dog food.
 
-`watch()` stays internal for framework code that needs it.
+`RouterView` should be rewritten as a compiled component that uses reactive JSX:
+
+```tsx
+// Before (imperative, uses watch)
+export function RouterView({ router, fallback }: RouterViewProps) {
+  const container = document.createElement('div');
+  watch(
+    () => router.current.value,
+    (match) => {
+      container.innerHTML = '';
+      if (match) container.appendChild(match.route.component());
+    },
+  );
+  return container;
+}
+
+// After (declarative, compiler-driven)
+export function RouterView({ router, fallback }: RouterViewProps) {
+  return (
+    <div>
+      {router.current.value
+        ? RouterContext.Provider(router, () => router.current.value!.route.component())
+        : (fallback ? fallback() : null)
+      }
+    </div>
+  );
+}
+```
+
+The compiler sees `router.current.value` in the JSX ternary, generates `__conditional` + `domEffect()`, and the route page swaps reactively. Disposal scope cleanup happens automatically (old page components get cleaned up by `__conditional`). No `watch()` needed.
+
+**Async/lazy routes:** The current `RouterView` hand-rolls `Promise.then` for lazy components. This should be a framework-level primitive (e.g., an `AsyncComponent` wrapper) that any component can use, not special-cased in `RouterView`.
+
+#### Framework components audit: eat your own dog food
+
+An audit of framework components found three that use imperative DOM patterns instead of the compiler-driven reactive flow:
+
+| Component | Anti-pattern | Fix |
+|-----------|-------------|-----|
+| **`RouterView`** (`router/router-view.ts`) | `watch()` + `innerHTML = ''` + `appendChild()` | Rewrite as compiled JSX component (see above) |
+| **`Link`** (`router/link.ts`) | `document.createElement('a')` + `effect()` + `classList.add/remove` | Rewrite as compiled JSX: `<a class={isActive ? activeClass : className} href={href}>{children}</a>` |
+| **`ThemeProvider`** (`css/theme-provider.ts`) | `document.createElement('div')` + loop of `appendChild()` + SSR VNode branch | Rewrite as compiled JSX: `<div data-theme={theme}>{children}</div>` |
+
+All three components were written before the compiler could handle their patterns. They should be refactored in Phase 2 to use declarative JSX compiled through the framework's own compiler — the same patterns we teach every user.
+
+**Why this matters:** These components are reference implementations. Developers read framework code to learn patterns. When `RouterView` uses `watch()` + `innerHTML`, developers copy that pattern instead of using reactive JSX. When `Link` uses `effect()` + `classList.add/remove`, developers think that's how reactive styling works instead of using class attributes. Framework code must demonstrate the intended usage patterns.
+
+Additionally, `element.ts` contains a leftover `console.log` debug statement in the `unwrapSignal` function that should be removed.
 
 #### No public escape hatch
 
-There is **no** `@vertz/ui/primitives` sub-export. `effect`, `signal`, `watch`, `domEffect`, `lifecycleEffect`, and `onCleanup` are not exported from any public entry point. Period.
+There is **no** `@vertz/ui/primitives` sub-export. `signal`, `domEffect`, `lifecycleEffect`, and `onCleanup` are not exported from any public entry point. `watch()` is removed entirely — zero callers after the RouterView refactor.
 
 Framework companion packages (`@vertz/ui-primitives`, `@vertz/tui`, `@vertz/ui-canvas`) are **first-party internal packages** — they import from relative paths within the monorepo, not from published package exports. They don't need a public sub-export; they have direct access to the source.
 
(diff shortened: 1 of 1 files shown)


viniciusdacal:
## Rev 2 — All Review Feedback Addressed

Updated the design doc to address feedback from all 5 reviews (3 agent reviews + 2 PR reviews). Here's what changed:

### Major additions

**1. Two-tier effect model** (resolves the `watch()`/`onMount()` SSR ambiguity — flagged by all reviewers)
- `effect(fn)` = DOM effect, **runs once during SSR** (populates text content, attributes, conditionals)
- `sideEffect(fn)` = side effect, **skips during SSR** (`watch()`, `onMount()`, `RouterView`)
- This is not a new bifurcation — it makes the existing implicit DOM-vs-side-effect distinction explicit and enforceable

**2. Error handling during SSR effects** (flagged by Design Review + Adversarial Review)
- DOM effects catch + log errors during SSR, don't crash the server
- Signal read-write cycles during SSR run exactly once (no tracking = no re-execution)
- Microtasks/timers in effects are fire-and-forget (same as every SSR framework)

**3. Two-tier export strategy** (flagged by DX Review)
- `@vertz/ui` — app developers (no `signal`, `effect`, `watch`)
- `@vertz/ui/primitives` — library authors (`ui-primitives`, `tui`, `ui-canvas`) who need raw primitives, with explicit opt-out of compiler optimizations

**4. Alternatives considered** (flagged by Adversarial Review)
- Compiler-level eager initialization — rejected (can't evaluate dynamic values at build time)
- `RenderMode` enum — rejected (adds param to every call, doesn't reduce the three paths)
- Marko-style dual compilation — rejected (doubles compiler output surface, new bug class)

### Fixes

- **Line count**: `ssrConditional` is 37 lines (lines 68–104), not 68. Fixed.
- **`__conditional` proposal**: Uses `parentNode.replaceChild()` (SSR shim compatible), not `node.remove()`. Full disposal scope management included.
- **Hydration**: `hydrateConditional` stays as an additive path for DOM claim. Explicitly verified in Phase 2 acceptance criteria.

### New sections

- **"What must NOT be deleted"** — router browser-API guards, `createRouter` SSR detection, `typeof document` guards
- **`onMount` SSR behavior** — explicitly documented as no-op, with disposal scope interaction explained
- **Expanded acceptance criteria** — `onMount`/`watch` skip SSR, error recovery, nested effects, signal read-write, performance baseline
- **Additional non-goals** — no compiler output changes, no server components, no Suspense in Phase 1–2

### Resolved unknowns

- Unknown #1 (cleanup during SSR): **No** — disposal scopes discarded with SSR tree
- Unknown #2 (`watch()` SSR behavior): **Skip** — uses `sideEffect()`, not `effect()`
- Reduced open questions from 6 to 4 (Phase 3+ only)

### Structural changes

- Bun migration research → **Appendix A** with explicit scope boundary ("separate effort, separate design doc")
- Breaking-change note at top: "greenfield project, breaking changes fully allowed"

### Dismissed concerns

Per VP Engineering note: all feedback about migration strategy, backward compatibility, and deprecation paths has been acknowledged and dismissed. The project has no external users.

viniciusdacal (commented):
## Design Review: Universal Rendering Model (Rev 2)

### Previous Concerns Addressed

The Rev 2 changes directly address feedback from the previous round:

1. **Two-tier effect model** - Resolves the SSR ambiguity for watch()/onMount(). The distinction between DOM effects (effect()) and side effects (sideEffect()) is clean and enforceable.

2. **Error handling during SSR** - Addresses the concern about effect failures crashing SSR. The try-catch + logging + placeholder approach matches industry standards (React, Solid).

3. **Two-tier export strategy** - Good DX solution. App developers get a simplified API; library authors can opt-in to primitives with explicit trade-off documentation.

4. **Alternatives considered** - Thorough coverage of why compiler-level fixes, RenderMode enum, and Marko-style dual compilation were rejected.

### Design Clarity: Ready for Implementation

The document is clear enough to implement:
- Phase 1 (two-tier effects): Clear API contract, acceptance criteria are testable
- Phase 2 (simplify primitives): Specific files and functions to delete are named
- Phases 3-6: Appropriately marked as future work with known unknowns

### Minor Gaps Clarifications Needed

1. Hydration mismatch detection (Phase 2) - This is mentioned but not specified. Will there be a schema/validation step, or purely DOM-structure comparison?

2. Signal write behavior during SSR effects - Consider: if a component renders twice, signal writes in the first effect could affect the second render. Is there a test case for this?

3. __conditional implementation - Is getIsHydrating() reliably set before the first DOM primitive call? What happens if hydration claim fails?

4. onMount cleanup edge case - If someone migrates from onCleanup() to onMount return, and calls onMount at module level, it silently fails. Is a dev-mode warning possible?

### Architectural Soundness

Strengths:
- Single source of truth for rendering logic
- Clean separation between DOM primitives and lifecycle hooks
- The effect/sideEffect distinction is principled
- Error resilience approach is sound

Concerns:
- The __conditional merged implementation is complex with nested scope management (pushScope/popScope) - needs careful testing
- Phase 1 acceptance criteria mention nested component rendering but dont specify depth/order verification

### Suggested Additions

1. Add to Phase 1: Test signal written in parent effect, read in child render (SSR order correctness)
2. Add to Phase 2: Test hydration claim failure falls back to CSR behavior

### Verdict

**Approved for implementation.** The design is sound, well-specified, and the two-tier effect model is the right architectural direction. Minor gaps identified are testable edge cases rather than fundamental design issues.

Issue #672, opened 2026-02-23
tracking: Component streaming (Phase 4)


viniciusdacal:
### Goal

Stream components progressively during SSR for faster time-to-first-byte.

**Design doc required before implementation.**

### Status: Future tracking issue — not yet planned for implementation.

viniciusdacal:
Closing — component streaming (Phase 4) is fully implemented.

Relevant commits:
- `b2f2c83` docs(design): component streaming for SSR [#672] (#683)
- `2e1d15e` feat(ui-server): implement component streaming for SSR [#672]
- `253d80c` fix(ui-server): address adversarial review findings in component streaming [#672]
- `389dbeb` docs(design): revise component streaming design after review [#672]

Implements data streaming via inline `<script>` tags with `window.__VERTZ_SSR_PUSH__()` API, CSP nonce support, XSS prevention, and error handling. Design doc at `plans/component-streaming.md`.

PR #683, opened 2026-02-24
docs(design): component streaming for SSR [#672]

viniciusdacal:
## Summary

Design doc for **Phase 4 of the Universal Rendering Model** (#664) — Component Streaming for SSR (#672).

### Key design decision: Data streaming, not HTML streaming

Instead of replacing HTML sections in the stream (fragile VNode tree surgery), the server streams **query data** via inline `<script>` chunks. The client's reactive system handles DOM updates automatically.

- **Fast queries:** Data baked into initial HTML (same as today)
- **Slow queries:** Loading placeholder in initial HTML → server streams resolved data → client swaps in automatically
- **Zero developer boilerplate** — `query()` handles everything
- **Global `ssrTimeout` default** with per-query override

### Architecture

```
Server: initial HTML → <script>__VERTZ_SSR_PUSH__("key", data)</script> → close
Client: hydration → CustomEvent listener → reactive signal update → DOM update
```

### Changes from current behavior

- `renderToHTML()` returns `Response` (streaming) instead of `Promise<string>`
- `renderToHTMLString()` added for backward compat
- Slow queries no longer require client-side refetch — server streams the data

## Review request

Looking for adversarial review on:
1. Is data streaming (vs HTML streaming) the right tradeoff?
2. Security implications of `<script>` injection for streamed data
3. Client-side hydration race conditions (data arriving before/after query init)
4. Global vs per-query timeout precedence rules
5. Breaking change impact of `renderToHTML` return type change

## Test plan

- [ ] Adversarial design review
- [ ] Approve or iterate on design before implementation

### DIFF
--- a/packages/ui-server/src/__tests__/render-to-html-stream.test.ts
+++ b/packages/ui-server/src/__tests__/render-to-html-stream.test.ts
@@ -287,6 +287,36 @@ describe('renderToHTMLStream', () => {
     expect(text).toContain('__VERTZ_SSR_PUSH__');
   });
 
+  it('cleanup runs after stream is fully consumed, not before', async () => {
+    const slow = deferred<string>();
+
+    function App(): VNode {
+      const item = query(() => slow.promise, {
+        key: 'stream-cleanup-timing',
+        ssrTimeout: 10,
+      });
+      const data = item.data as unknown as { value: string | undefined };
+      return {
+        tag: 'div',
+        attrs: { id: 'app' },
+        children: [data.value ?? 'Loading...'],
+      };
+    }
+
+    const response = await renderToHTMLStream({ app: App, url: '/', ssrTimeout: 10 });
+
+    // At this point, the Response has been returned but the stream is NOT consumed yet.
+    // The global ssrTimeout should NOT have been cleared yet (because the stream is still open).
+    // Resolve the slow query — this should succeed because cleanup hasn't run yet.
+    slow.resolve('Cleanup Test Data');
+
+    const text = await response.text();
+
+    // The stream should have completed successfully
+    expect(text).toContain('Loading...');
+    expect(text).toContain('Cleanup Test Data');
+  });
+
   it('failed slow query does not crash, remaining queries still stream', async () => {
     const failing = deferred<string>();
     const succeeding = deferred<string>();
--- a/packages/ui-server/src/__tests__/ssr-context.test.ts
+++ b/packages/ui-server/src/__tests__/ssr-context.test.ts
@@ -1,9 +1,11 @@
 import { describe, expect, it } from 'vitest';
 import {
+  clearGlobalSSRTimeout,
   collectSSRError,
   getSSRErrors,
   getSSRQueries,
   registerSSRQuery,
+  setGlobalSSRTimeout,
   ssrStorage,
 } from '../ssr-context';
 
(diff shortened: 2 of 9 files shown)


viniciusdacal (commented):
## Design Review: Component Streaming for SSR

Overall this is a solid design that addresses a real problem. The data-streaming-via-script approach is pragmatic and avoids the fragility of HTML replacement. A few areas need attention before implementation:

### Clarity & Completeness

**✅ Good:**
- Clear problem statement and context
- Implementation steps are well-organized
- File table is helpful
- Backward compat via `renderToHTMLString()` is smart

**⚠️ Needs clarification:**
- **Step 2**: "don't()" — clarify block on response.text how initial HTML is sent. Does `renderToHTML()` return a Response that starts streaming immediately, or do we need to change the caller to handle streaming?
- **Key generation**: The design assumes `cacheKey` is deterministic and consistent between SSR and client. What's the format? How is it generated to ensure stability?

### Architectural Soundness

**✅ Solid choices:**
- Script-tag data injection avoids VNode surgery
- CustomEvent pattern is clean for cross-boundary communication
- Global timeout hook follows existing `__VERTZ_IS_SSR__` pattern

**⚠️ Concerns:**

1. **Race condition risk**: If client hydration starts and `query()` checks for SSR data *before* the streamed script arrives, it will refetch. The client needs to wait for `vertz:ssr-data` event before deciding to refetch. Current Step 6 doesn't address this — it just checks `window.__VERTZ_SSR_DATA__` immediately.

2. **Memory leak potential**: The `window.__VERTZ_SSR_DATA__` array accumulates all streamed data and is never cleaned up. With many slow queries or long sessions, this grows unbounded.

3. **Missing SSR detection**: How does client-side code know it's running on an SSR-rendered page? The design checks `window.__VERTZ_SSR_DATA__` but that's created by the runtime script. What if SSR failed or was disabled? Need explicit `__VERTZ_SSR_ENABLED__` flag.

### Edge Cases

**Missing handling for:**

1. **Query resolves AFTER client-side fetch started**: If the client starts refetching before streamed data arrives, we could have a race. Consider deduplication or cancellation.

2. **Partial query failures**: What happens when some queries in a batch succeed and others timeout? Error state isn't propagated to client.

3. **ssrTimeout: 0 behavior**: Document says "no streaming (current behavior)" — but what exactly happens? Does it show loading forever? Client refetch? Need explicit contract.

4. **Stale data**: No max-age or cache headers on streamed responses. CDNs may not cache streaming responses properly.

5. **Multiple renders**: What if `renderToHTML()` is called multiple times in the same request (e.g., for redirects)? Need to ensure clean state.

### Acceptance Criteria

Current verification items are good but add:
- `query()` with pending SSR data should NOT make network request (verify in network tab)
- Memory usage stays flat after hydration (no growing array)
- Failed queries don't crash the stream — remaining queries still stream

### Nitpicks

- **Naming**: `__VERTZ_SSR_PUSH__` suggests push from server, but it's actually server→client via script. Consider `__VERTZ_SSR_DATA__` as the bus name to match the array.
- **Error types**: No custom error types for SSR-specific failures (timeout, hydration mismatch, etc.)
- **TypeScript types**: Need public types exported for `QueryOptions.ssrTimeout`

### Questions

1. How does this interact with query deduplication/coalescing? 
2. Is there a limit on how many queries can be streamed per request?
3. What's the hard timeout for "all queries resolve or hard timeout" in Step 2?

Overall: Good foundation, address the race condition and memory concerns before coding.

